PR #44, opened 2015-10-08
Add utility function createBlankWindow and refactor the code to use that, make default window width and height to be minWidth and minHeight. fix window menu not showing open dialogs.

aminroosta:


aminroosta:
@arnabk did you fixed jquery.extended installation? (check my messages on gitter)

PR #60, opened 2015-10-16
WS has a bug, when an error accures no echo_req will be returned. thi…

aminroosta:
…s needs to be fixed, We cant tell which promise should be rejected. As a TEMPORARY WORKAROUND because requests are returned in FIFO order(not documented), we will reject the first one


aminroosta:
This **will** be fixed from server side, so i will close this.

PR #73, opened 2015-10-27
windows.js has hard coded instrument list which gets loaded on page load. Fix it. Should load randomly from the list getting loaded in instruments.js call of trading_times

aminroosta:


arnabk:
@amin As far as I know delay_item is in feed. Please fix this.
delay_amount: 0 //TODO fix this when API provides it
I will hold on to your PR until this is fixed.

PR #105, opened 2015-11-08
Refactoring code. Simplifying sortMarkets function. Make sure that chartonly filter is used by all trading_times callers

arnabk:


src/websockets/binary_websockets.js
```diff
@@ -72,12 +72,22 @@ define(['es6-promise', 'reconnecting-websocket', 'js-cookie', 'token/token', 'jq
...
             if (data.error)
                 promise.reject(data.error);
+            else if (data.echo_req.trading_times) {
+                require(['common/menu'], function(menu) {
+                    markets = menu.extractMenu(data, {
+                        filter: function (sym) { return sym.feed_license !== 'chartonly'; }
+                    });
+                    promise.resolve(markets);
+                });
+            }
```
aminroosta: I think its not a good idea to pre-process data in **binary_websockets.js**, 
this is not consistent, If we what to do that, its better to pre-process every request.
i think every single caller of **{trading_times: ...}** should call *menu.extractMenu()* in order to get the **markets** in the format it wants.

aminroosta: what if some other module does not need the filter ** sym.feed_license !== 'chartonly'; ** ?

PR #119, opened 2015-11-11
Add the feature to deploy branches to sub-folders of gh-pages.

aminroosta:
For example, a branch named aminassetindex will be deployed to aminroosta.github.io/webtrader/aminassetindex. which means multiple branches can be deployed to subfolders of gh-pages at the same time.

@arnabk this PR has the changes from #114 (but since you reverted that in #115 it does not show those changes), I think there will be no git problems, if you merge that one first.


arnabk:
@aminroosta Can you also finish this task in this PR?
https://trello.com/c/EfloHSqW/165-implement-grunt-gh-pages-clean-task

aminroosta:
Done.

PR #120, opened 2015-11-11
Init WS as the first thing when page loads. All others can continue to work parallely. This will help WS to be ready by the time other items finish download.

aminroosta:
@arnabk Deployed to aminroosta.ir/webtrader/liveapi2
Check the chrome developer network tab, you can see that **binary_websocket.js** is loaded among the first scripts in less that 50ms.


aminroosta:
@arnabk merge this after reverting #115

PR #134, opened 2015-11-30
Add Bollinger Bands(BBANDS) indicator

MahboobehMohammadi:
These are some questions related to this task:
1- Is Bollinger Band only an OHLC indicator?
2 -This Indicator has three bands,I set default colors for them of them, should each of them has it's own color picker or setting default color is fine?
3- What is max and min period for this indicator? It has been considered from 1 to 200 in this commit.
4- This indicator has three series,I'm not sure if each series needs different unique IDs or not?and why?


arnabk:
@MahboobehMohammadi 
1. No. Its for all types
2. Color picker should be provided
3. We have to go by standards. We have to provide that restriction. A meaningful value would be 1-50 for time period, 1-20 for Dev up and Dev Dn. Anything that makes the chart look odd should not be allowed. Most of the time the value selected by users are 20, 2, 2
4. Any series in highcharts need unique ID. Failing to do so would create issues to take them out when user selects to remove indicator.

arnabk:
Questions - 
> Open a tick chart and add BB indicator. BB is not updating with realtime charts.

> Color choice should be provided to the user. Check other indicator code.

> Ideally users should be give these inputs

      > Price to select. OPEN, HIGH, LOW, CLOSE

      > Time period

      > Dev up

      > Dev Dn

      > MA type(EMA, SMA, WMA etc)

> When we remove, the indicator is leaving 2 of the series lines(out of 3)

> Are you sure this code is not going to work when you load BB in 2 different charts with different parameters? You are storing variables which will be shared among charts.

> There is a copy paste error H.Series.prototype.addBOP You still have this line in code

> Can we simplify and optimize the code?

MahboobehMohammadi:
1- I don't know what is Dev up, Dev Dn and MA type(EMA, SMA, WMA ),Can you please tell me which other charts use these, then I can check them out?
2- Yes I think it's because of uniqueIDs, I"ll chack it out.I'ill fix the removing and updating parts and i'll also add color pickers,stroke width and dash style for each bands.
3- Yes,sure.I will optimize it.

arnabk:
I think you got your answers in email. Please default the values for period as 12, then Dev up as 2 and Dev Dn as 2. Check the screenshot
screen shot 2015-12-01 at 7 21 34 am

arnabk:
For review comments
https://github.com/binary-com/webtrader/pull/141

arnabk:
@MahboobehMohammadi Everything looks good. 
But one issue.
Steps - 
1. Open R_25 tick chart
2. Add BB indicator
3. Then open remove indicator screen
You will note that there are 3 BB indicator names shown.
This is happening because all of them has isIndicator variable true.

Load BB on R_25 1Min and you will see that the values are not changing on the current forming bar
Also the values are not rounded to 4 decimal places.

MahboobehMohammadi:
@arnabk I rounded values to 4 decimal places,but i couldn't see the "values are not changing on the current forming bar" problem,Would you please send me a screen shot of the problem?

arnabk:
@MahboobehMohammadi Watch the video. There is some issue with the update of BB
https://drive.google.com/file/d/0BxkTYvth5JGWc1lEb191WnBycXc/view?usp=sharing

PR #147, opened 2015-12-18
Cci

arnabk:


arnabk:
@MahboobehMohammadi Here are comments
* Rename ’Ok’ to ‘OK’
* Why do we need getPrice? Can we update extractPriceForAppliedTO to do what getPrice is supposed to do? You have to refactor the code and make sure that wherever you used getPrice are replaced with this method
* What is ma_type_service.js? Do we really need it when everything has been moved into indicator_base.js?
* Please update tema, ema, sma, trim, macd, wma, bbands indicators to use what you coded in indicator_base.json

MahboobehMohammadi:
- I'm using getPrice in different situations,when data is just an array of prices ,when it contains indexes and prices and when it's an OHLC and needs extractPriceForAppliedTO to extract the price,and these two,getPrice and extractPriceForAppliedTO ,are being used in different places,I don't know if it's causing any problems to combine these two?and what about extractPrice?

- ma_type_service has been removed.

- tema, ema, sma, trima, macd, wma, bbands indicators has been updated.

arnabk:
* I think your getPrice is same as extractPrice. We definitely have to get rid of getPrice in some way. I will look into this later.
* Please down merge and update this fork. It has merge conflicts.

MahboobehMohammadi:
I update it ,I didn't have any conflict.

PR #148, opened 2015-12-20
Implement Login Logout buttons in navigation bar

aminroosta:
@arnabk  deployed to aminroosta.ir/webtrader/aminlogin


arnabk:
@aminroosta Please review this method. I do not think this is correct. `arguments` variable does not exist. Why are we using setTimeout? I am not sure if you need fireEvent method. What is wrong if we just have a method and call synchronously?

    /* fire a custom event and call registered callbacks(api.events.on(name)) */
     var fire_event = function(name /*, args */){
       var args = [].slice.call(arguments,1);
       var fns = callbacks[name] || [];
       fns.forEach(function (cb) {
           setTimeout(function(){
             cb.apply(undefined, args);
           },0);
       });
     }

aminroosta:
@arnabk  arguments is a special object in js functions [mozilla docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments)

I am using **setTimeout** because i don't want to block the caller.  the callbacks will run async. and if one of them throws an exception, it will not break the caller code.

The purpose of **fire_event** is to take a **name** argument and call all of the subscribed handlers for that with the rest of its arguments.

@arnabk you said **Design also says that it has to be extreme right**  i don't understand what you mean ?

And the Login button being disable forever, was a bug :) i have fixed it now with #d7e4077.

arnabk:
@aminroosta Aha, I did not know about that arguments parameter. 

Move this to the extreme right top corner of the page. Thats where the login button and the logged in yours ID belongs to.

aminroosta:
@arnabk  styles have been updated. please take another look at aminroosta.ir/webtrader/aminlogin :smiley:

PR #157, opened 2015-12-24
Statement dialog live scroll.

aminroosta:
@arnabk i have tried a couple of different libraries for live scroll, but no lock!
I was going to give up on this task :tired_face: 

I tried and implemented a live scroll, my implementation very basic :
when the user scrolls more 75%, then i will request 50 more rows and add them to the table.
If a specific date is selected then live scroll is disabled and only the records for that specific date will be shown.
If the user clears the selected date again, then the normal live scroll will work again.


aminroosta:
deployed to aminroosta.ir/webtrader/aminstatement

arnabk:
@aminroosta 
* If a specific date is selected then live scroll is disabled and only the records for that specific date will be shown.
What if that specific date has got thousands of records?

PR #176, opened 2016-01-05
Correct use of active_symbols api in trade menu

aminroosta:


aminroosta:
@arnabk deployed to aminroosta.ir/webtrader/amintrading

PR #183, opened 2016-01-12
Debug ind

MahboobehMohammadi:


src/charts/chartingRequestMap.js
```diff
@@ -213,13 +215,11 @@ define(['lokijs', 'lodash', 'jquery', 'websockets/binary_websockets', 'common/ut
             var req = {
                 "ticks_history": options.symbol,
                 "granularity": granularity,
+                "subscribe": options.subscribe || 0,
                 "count": options.count || 1,
                 "end": 'latest',
                 "style": style
             };
-            if (options.subscribe === 1) {
```
arnabk: We cannot remove this. Let's discuss. We cannot send subscribe : 0 . Server will throw error for delayed instruments. Let's discuss why you need it back.

PR #191, opened 2016-01-19
Implement live scrolling for profit_table

aminroosta:


src/main.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>Binary.com : Web Trader</title>
+    <title>Binary.com : Webtrader</title>
```
arnabk: Is this supposed to be one word?

aminroosta: Yes, parashuram asked me on slack to change it.

PR #251, opened 2016-02-08
Do not miss ticks when shwoing tick live confirmation.

aminroosta:
I was not able to reproduce the issue directly, but after manually delaying creation of confrim dialog i saw that some ticks are missing.
this pr tries not to miss any previous ticks by checking *barstable* first.


aminroosta:
deployed to aminroosta.ir/webtrader/amintrading.

PR #268, opened 2016-02-13
Fixed Rising/Falling Three Methods

MahboobehMohammadi:


arnabk:
@MahboobehMohammadi The implementation is not correct. This pattern can be formed with 3 or 7 candles. Please the definition again in http://thepatternsite.com/

PR #274, opened 2016-02-17
Fix indicator validations and patterns broken images in IE

MahboobehMohammadi:


src/common/util.js
```diff
@@ -51,6 +51,11 @@ function isNumericBetween(value, min, max) {
     return isNumeric && Math.floor(value) == value && min <= value && max >= value;
 };
 
+function isFloatBetween(value, min, max) {
```
arnabk: @MahboobehMohammadi 
Revert this change and use https://lodash.com/docs#inRange
Lodash is already in our project. You have to include it properly in the project using requireJS for the indicators where you want to use them. In your update PR, this file should not show up.

PR #280, opened 2016-02-20
Add Fractal Indicator

MahboobehMohammadi:
@arnabk Could you please take a look at its AddPoint and Update.The code seems to be correct but its not working as it suppose to.I might missed something.The difference between fractal and other indicators is that the others use previous candles to find a pattern,this one uses next candles.so when a candle is being added or updated,it influence its previous candle pattern.


arnabk:
- I changed code where I think was not correct. Please review it and let me know if everything is working fine. 
- Add validation to make sure that the input value is always odd. We cannot accept even number for this input.
- Do not show hover over X and Y values on fractal points
- In line charts, y = close value
- In candle or OHLC charts, y = high/low (depending on bear or bull)
- Test live mode. That is, load it on R_25, 1Minute chart and test it for couple of minutes to see if the markers are showing up when the pattern is there, and getting removed, when the pattern is invalidated. This can only happen at the end of the chart on right side. You have to watch it.
- Also test few CDL indicators to make sure that my changes are not breaking them since I tried to reuse common code

Other these changes, everything else looks fine.

MahboobehMohammadi:
I tested it and made some fixing but the live mode has problem.It doesn't work properly.The other things you mentioned have  been done.

PR #282, opened 2016-02-21
Add percentage change to chart Options > Chart Type > Table

MahboobehMohammadi:


arnabk:
@aminroosta Please review

aminroosta:
@MahboobehMohammadi Please deployed your changes to gh-pages.

src/charts/tableView.js
```diff
@@ -12,14 +12,16 @@ define(['jquery', 'moment', 'lokijs', 'charts/chartingRequestMap', 'websockets/s
     table.animate({left: '0'}, 250);
     chart.animate({left: '-100%'}, 250);
     refresh_table(dialog, key); /* clear table and fill it again */
+    dialog.parent().css('width', 520);
```
aminroosta: Please only increase the width of dialog when the table width is bigger than the dialog.
And the content of table does not fit.
Also you may consider using a jquery animation :)

src/charts/tableView.js
```diff
@@ -12,14 +12,16 @@ define(['jquery', 'moment', 'lokijs', 'charts/chartingRequestMap', 'websockets/s
...
     dialog.view_table_visible = true; /* let stream_handler new ohlc or ticks update the table */
   }
   var hide_table_view = function(dialog){
     var table = dialog.find('.table-view');
     var chart = dialog.find('.chart-view');
     table.animate({left: '100%'}, 250);
-    chart.animate({left: '0'}, 250);
+    chart.animate({ left: '0' }, 250);
     dialog.view_table_visible = false;
+    dialog.parent().css('width', 350);
```
aminroosta: You should set the width to previous dialog size, the user might have resized the dialog.
Or it might be in fullscreen.

MahboobehMohammadi: @aminroosta Done!

src/charts/tableView.js
MahboobehMohammadi: @aminroosta The size of table is not changing until the dialog width changes,so I'm using  this constant value that is needed to avoid horizontal scroll.

PR #283, opened 2016-02-21
Change ind levels and fix AO validation and change AO dialog size

MahboobehMohammadi:


arnabk:
@MahboobehMohammadi Can you please move level.css, level.html and level.js to indicators folder instead of creating a folder called "lavel". A folder means an indicator, which could be mis-leading here

MahboobehMohammadi:
@arnabk The level files have been moved to the indicators folder and I added the path in main.js.

PR #285, opened 2016-02-25
Test and debug percentage column

MahboobehMohammadi:


arnabk:
@MahboobehMohammadi Chart > Table is very slow as compared to previous version of beta

src/websockets/stream_handler.js
```diff
@@ -38,7 +38,11 @@ define(["websockets/binary_websockets", "charts/chartingRequestMap", "common/uti
                 }
                 barsTable.insert(tick);
                 /* notify subscribers */
-                var preTick = barsTable.data[barsTable.data.length - 2];
+                var preTick = tick;
+                var bars = barsTable.chain().find({ instrumentCdAndTp: key }).data();
```
arnabk: @MahboobehMohammadi This is wrong. You should never get all data if you just one bar. Read about lokijs. Check other code. Always optimize your code. Use

> barsTable
>                                 .chain()
>                                 .find({instrumentCdAndTp: key})
>                                 .where(function (obj) {
>                                     return obj.time >= lastBarOpenTime;
>                                 })
>                                 .simplesort('time', false)
>                                 .limit(1).data();

src/websockets/stream_handler.js
```diff
@@ -109,11 +113,11 @@ define(["websockets/binary_websockets", "charts/chartingRequestMap", "common/uti
                 barsTable.update(bar);
             }
             
-            var preOhlc = bar;
-            if (isNew && barsTable.data.length > 1)
-                // get the previous bar
-                preOhlc = barsTable.data[barsTable.data.length - 2];
             /* notify subscribers */
+            var bars = barsTable.chain().find({ instrumentCdAndTp: key }).data();
```
arnabk: @MahboobehMohammadi This is wrong. You should never get all data if you just one bar. Read about lokijs. Check other code. Always optimize your code. Use

barsTable
.chain()
.find({instrumentCdAndTp: key})
.where(function (obj) {
return obj.time >= lastBarOpenTime;
})
.simplesort('time', false)
.limit(1).data();

PR #289, opened 2016-03-01
Change ind validations to allow last number in range and also check for integer values

MahboobehMohammadi:


arnabk:
This has merge conflict

MahboobehMohammadi:
@arnabk I can't see any conflict and It has been updated from development branch,what else can i do?

PR #297, opened 2016-03-04
Remove hamburger.css

MahboobehMohammadi:


MahboobehMohammadi:
@arnabk I removed the hamurger.css and it's path in main.js but i have to keep the styles that have been used in chart options unless you want to change the style itself and use a simple menu like navigation menus because these styles(the hamburger menu(the three black line)) are nowhere else in the code.

MahboobehMohammadi:
@arnabk The style that is being used in top navigation menu can not be used everywhere, It has been created for a specific situation, and it get messy if we want to use it here because some of it's behaviours are not necessary everywhere so It's much better if we just add a simple style that can be use anywhere we need hamburger menus.

arnabk:
@MahboobehMohammadi Adding my comment in trello

PR #303, opened 2016-03-08
Fix responsive trading link at the bottom of page

MahboobehMohammadi:


src/windows/windows.js
```diff
@@ -455,6 +456,10 @@ define(['jquery', 'lodash', 'navigation/navigation', 'jquery.dialogextend', 'mod
             closeAllObject && closeAllObject.click();
         },
 
+        fixFooterPostion:function()
```
arnabk: Can we do fixFooterPostion : fixFooterPostion ?

MahboobehMohammadi: @arnabk Done.

PR #308, opened 2016-03-11
Set expiration date for webtrader_token cookie

MahboobehMohammadi:
@arnabk We are using js-cookie to set the cookie and by default it removes the cookie when the user closes the browser, but we can define when the cookie will be removed. Value can be a number which will be interpreted as days from time of creation or a date instance, that i set it to 1 day.


arnabk:
@MahboobehMohammadi Make it never expires. The original plan was to never expire it.

MahboobehMohammadi:
@arnabk Done.

PR #320, opened 2016-04-05
Drawing tools circle

ankitbinary:
Added Files for Base drawing Constructor
Added files for drawing Circle (Dialog option files and Sub Class file )


arnabk:
Its not working. See the error message.

screen shot 2016-04-05 at 1 52 12 pm

Also I cannot merge this PR because you have enabled all other drawing tools, but this PR only submits changes for Circle. Anything makes to the upstream has to be perfect

PR #321, opened 2016-04-07
Historical data dialog

ankitbinary:
Fixed the textual changes from 15min to 15mins
Fixed the View Historical data Dialog Time period


arnabk:
Invalid PR

PR #326, opened 2016-04-08
Added decimals to Trade ==> Dialog == > Return % , for Accuracy

ankitbinary:
Added two decimals for 'return percent' value on Trade Dialog


arnabk:
Your changes are almost near to what actually needed, but its not completely correct. I am closing this as I fixed in the binary-com/webtrader:development branch

PR #330, opened 2016-04-10
Added  3 Digit decimals on trade chart tooltip

ankitbinary:
Added 3 Digit decimals on trade chart tooltip

### DIFF
--- a/src/viewtransaction/viewTransaction.js
+++ b/src/viewtransaction/viewTransaction.js
@@ -52,7 +52,7 @@ define(["jquery", "windows/windows", "websockets/binary_websockets", "portfolio/
           text: title,
           style: { fontSize:'16px' }
         },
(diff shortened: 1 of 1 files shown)


arnabk:
4 decimal contracts are working without your changes.
image

Why do we need this change for 3 decimal contracts? I am worried that your change will break other contracts. We cannot show 3 decimals for those which are 2 decimals or 4 decimals. If Random 100 is working without your changes, then Random 25 should also work without this change. There might be issue with code somewhere else.

arnabk:
@ankitbinary It is not straight forward to conclude that the instrument could be formatted with 3 digits. Some instruments require 3, others 2 and 4. Its better we get this information from backend. Right now, we are getting this information in https://developers.binary.com/api/#active_symbols. However, its better to get the same "pip" information in https://developers.binary.com/api/#trading_times so that we could use it for this as well as for all other charts.

I have opened a defect with backend team in order to track this dependency.

PR #343, opened 2016-04-20
Final price shifting

ankitbinary:
Fixed the price shifting it's value


arnabk:
This change might break other trade types.

PR #344, opened 2016-04-21
Fixed the YAxis Title Align

ankitbinary:
Fixed the Aligining of the Axis Title , and it will not overlap with the Axis line


arnabk:
@ankitbinary This is going to break all other indicators. Modifying IndicatorBase will effect all indicators. Please check my changes here https://github.com/binary-com/webtrader/commit/e33302097eb9a3f08edff3d8d34aa576cd4879f7

PR #348, opened 2016-05-02
remove cached promises on logout

aminroosta:
screen shot 2016-05-02 at 11 33 44 am

Only `asset_index` and `trading_times` are cached right now.
Do we need to clear the cache?


aminroosta:
@arnabk  updated. now only cached authorized requests will be removed on logout.
deployed to aminroosta.ir/webtrader/aminliveapi

PR #360, opened 2016-05-11
Fix getCurrentPriceOptions

4p00rv:
This fix is for [trello card](https://trello.com/c/nccGbdTb/509-toggling-current-price-has-no-effect).


arnabk:
https://github.com/binary-com/webtrader/commit/8a62a35795d39b8fe363a1b803e007c84dd9da5e will fix the issue. This PR is not needed

PR #386, opened 2016-05-25
Shashank/Fixing clickjacking

Shashank-In:
Removing frameable response to avoid clickjacking attacks on the webtrader

### DIFF
--- a/src/main.html
+++ b/src/main.html
@@ -4,6 +4,7 @@
     <meta charset="UTF-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta http-equiv="X-Frame-Options" content="SAMEORIGIN">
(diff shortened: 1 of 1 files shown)


borisyankov:
binary-static uses iframe to embed a chart from webtrader.

arnabk:
@Shashank-In @borisyankov Will this break binary-static?
Our affiliates might also include this chart in frame. As far as I know Marketsandyou use it in their frame.

Shashank-In:
@arnabk @borisyankov Yeah that will break the static . Thou we can whitelist them . That would be simple fix . But the thing is later I found that this fix can be bypassed . So using meta to fix the bug is useless. We have to make server side fix which is not possible in github pages I guess. Please dont not merge . Wait for some more time . Else I would mail seniors to help me out

arnabk:
@Shashank-In Please resubmit a new PR or reopen this once you are sure if it is not going to break our chart embedding in IFRAME for both affiliates and binary-static

PR #390, opened 2016-05-26
Add lines to drawing tools

4p00rv:
Added horizontal and vertical lines

### DIFF
--- a/src/charts/draw/chartobject_remove.js
+++ b/src/charts/draw/chartobject_remove.js
@@ -102,29 +102,10 @@ define(["jquery", "datatables", 'charts/charts'], function ($) {
         },
 
         showDialog : function(containerIDWithHash){
-            var chart = $(containerIDWithHash).highcharts(),
-                middle = chart.chartWidth/2 - 130,
-                text = "Double click the object to remove.";
+            require(["jquery", "jquery-growl"], function($){
+                $.growl.warning({message:"Double click the object to remove."});
+            });
             
-            chart.display = chart.renderer.label(text, chart.plotLeft + middle, chart.plotTop)
-                .attr({
-                    fill: Highcharts.getOptions().colors[0],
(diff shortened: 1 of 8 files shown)


arnabk:
@4p00rv Can you move the cards which this task relates too?

src/charts/draw/chartobject_remove.js
```diff
@@ -99,8 +99,33 @@ define(["jquery", "datatables", 'charts/charts'], function ($) {
...
+                    r: 5,
+                    zIndex: 8,
+                    opacity: 0.95
+                })
+                .css({
+                    color: '#FFFFFF'
+                })
+                .add();
 
+            setTimeout(function () {
```
arnabk: @4p00rv What exactly are you trying to achieve by delaying this execution by 2 seconds?

src/charts/draw/chartobject_remove.js
```diff
@@ -99,8 +99,33 @@ define(["jquery", "datatables", 'charts/charts'], function ($) {
...
+                chart.display.fadeOut();
+            }, 2000);
+            setTimeout(function () {
+                chart.display = chart.display.destroy();
+            }, 2500);
```
arnabk: @4p00rv Same question here

4p00rv: @arnabk The faded message should be removed from the chart.

src/charts/draw/chartobject_remove.js
4p00rv: @arnabk I'm just displaying a message that fades away in 2 seconds. Since the message shouldn't be always displayed on the chart (to avoid this message from interfering with the view). I'am removing it in the next step.

4p00rv:
@arnabk I have added growl warning and icons.

PR #405, opened 2016-06-10
fix ohlc

4p00rv:
Fix for ohlc bar not getting updated properly.

### DIFF
--- a/src/websockets/stream_handler.js
+++ b/src/websockets/stream_handler.js
@@ -2,14 +2,6 @@ define(["websockets/binary_websockets", "charts/chartingRequestMap", "common/uti
 
     var barsTable = chartingRequestMap.barsTable;
 
-    function setExtremePointsForXAxis(chart, startTime, endTime) {
-        chart.xAxis.forEach(function (xAxis) {
-            if (!startTime) startTime = xAxis.getExtremes().min;
-            if (!endTime) endTime = xAxis.getExtremes().max;
-            xAxis.setExtremes(startTime, endTime);
-        });
-    }
-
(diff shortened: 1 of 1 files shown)


src/websockets/stream_handler.js
```diff
@@ -136,31 +128,28 @@ define(["websockets/binary_websockets", "charts/chartingRequestMap", "common/uti
...
                     if (isNew) {
                         series.addPoint([time, close], true, true, false);
                     } else {
-                        last.update({
-                            y : close
-                        });
+                        last[1] = close;
+                        chart.series[0].options.data[len] = last;
+                        // Redrawing the chart to update new point.
+                        chart.series[0].setData(chart.series[0].options.data, true);
```
arnabk: @4p00rv Is this going to be called for every update?

4p00rv: @arnabk Yes, that's what the developer suggested as a fix.

arnabk: can you send me the link. This seems to be very bad suggestion. For every update, it is going to reset so many things and refresh all data.

4p00rv: Here's the link: https://github.com/highcharts/highcharts/issues/5385
Its the second last comment in which the developer suggests to use `series.setdata`, although he seems to be concerned about performance too.

arnabk:
@4p00rv I looked at it. Its a bit risky change. We have bunch of indicators coded and they will not update if we change values like this. Highchart data update needs to trigger when we update data point in order to trigger calculation in indicators

PR #406, opened 2016-06-12
Add css property selector for plotLines

4p00rv:


arnabk:
@4p00rv Reduce the size of the dialog. its too big

4p00rv:
@arnabk I have updated the pr.

PR #412, opened 2016-06-20
Custom theme

4p00rv:
This pr enables a user to create custom theme.


src/charts/chartWindow.js
```diff
@@ -42,7 +42,6 @@ define(["jquery","windows/windows", "text!charts/chartWindow.html", 'lodash', "j
                 },
                 resize: _trigger_Resize_Effects
             }, options );
-
```
arnabk: @4p00rv Looks like this file change is not needed. Please revert. 
Its better to submit files which really has some change so that I get an idea of how many files you touched for particular task. This gives me some idea of impact of such PR

src/charts/charts.js
```diff
@@ -165,7 +165,6 @@ define(["jquery","charts/chartingRequestMap", "websockets/binary_websockets", "w
          */
         drawChart: function (containerIDWithHash, options, onload) {
             var indicators = [];
-
```
arnabk: @4p00rv Looks like this file change is not needed. Please revert. 
Its better to submit files which really has some change so that I get an idea of how many files you touched for particular task. This gives me some idea of impact of such PR

src/charts/charts.js
```diff
@@ -188,7 +187,6 @@ define(["jquery","charts/chartingRequestMap", "websockets/binary_websockets", "w
                 type : options.type,
                 delayAmount : options.delayAmount
             });
-
```
arnabk: @4p00rv Looks like this file change is not needed. Please revert. 
Its better to submit files which really has some change so that I get an idea of how many files you touched for particular task. This gives me some idea of impact of such PR

src/themes/themes.js
```diff
@@ -6,12 +6,19 @@ define(['jquery', 'windows/windows', 'highstock', "jquery-growl"], function($, w
...
     themeName = themeName && themeName.name;
     if (themeName) {
         require(['lib/highstock/themes/' + themeName]);
+    } else if(custom_theme){
+        console.log(custom_theme);
+        Highcharts.setOptions(custom_theme);
     }
 
+    //For custom theme.
+    require(['themes/custom_theme/custom_theme']);
```
arnabk: @4p00rv This is incorrect. You require where you need. You just should not require something if you are not going to use it.

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+				temp = temp + "{" + "\"" + property + "\"" + ":" + "\"rgba(" + rgba.join(", ") + ")\"";
+				return;
+			}
+			temp = temp + "{" + "\"" + property + "\"" + ":";
+		});
+		temp = temp + end;
+		$.extend(themeConf, JSON.parse(temp));
+	}
+
+	$('a.theme_custom')
```
arnabk: @4p00rv This is incorrect. Please refer to download.js and navigator.js how it is getting loaded on click of the menu item. We should load modules when users want it

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+    .off('click')
+    .on('click', function(){
+
+    	var ele = $(this);
+
+    	require(['css!themes/custom_theme/custom_theme.css']);
```
arnabk: @4p00rv You are requiring this every time I am clicking the menu. Not good.

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+            closable: false,
+            closeOnEscape: false,
+            width: 650,
+            height: 515,
+            title: "Customize chart appearance",
+            modal: true,
+            my: 'center',
+            at: 'center',
+            of: window,
+            destroy: function() {
```
arnabk: @4p00rv I think you dont need my,at, of. It should show in center by default.

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+            buttons: {
+              Apply: function() {
+              		openConfirmDialog();
+              },
+              Cancel: function() {
+                  $( this ).dialog( 'close' );
+                  $( this ).dialog( "destroy" );
+                  var themeName = local_storage.get("theme");
+                  themeName = themeName && themeName.name;
+                  //Reseting charts to previous theme or default.
```
arnabk: @4p00rv Do you really need to do this?

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+            }
+      		});
+
+    		win.dialog('open');
+    		generateData();
+    		createChart();
+      });
+  });
+
+  function generateData() {
```
arnabk: @4p00rv Instead of random numbers, please hard code some nice OHLC code. You can get it easily from our developers.binary.com. Just get 10-20 values. We need candles and not line

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+          }
+      },
+
+      title: {
+          //Show name on chart if it is accessed with affiliates = true parameter. In normal webtrader mode, we dont need this title because the dialog will have one
+          text: "Some random index"
+      },
+
+      credits: {
+          href: 'http://www.binary.com',
```
arnabk: @4p00rv Get the code from charts.js This href will navigate away from webtrader.binary.com

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+	            height: 240,
+	            resizable: false,
+	            collapsable: false,
+	            minimizable: false,
+	            maximizable: false,
+	            closable: false,
+	            closeOnEscape: false,
+	            modal: true,
+	            ignoreTileAction:true,
+	            'data-authorized': 'true',
```
arnabk: @4p00rv This is incorrect. This is not a dialog which would require users to be under authorized context.

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+	            destroy: function() {
+	                win = null;
+	            },
+	            buttons: {
+	                Apply: function() {    
+		                  $.growl.notice({message: 'Loading custom theme.'});
+		                  local_storage.remove('theme');
```
arnabk: @4p00rv In order to use local_storage, you should require 'common/util' first

src/themes/custom_theme/custom_theme.js
```diff
@@ -0,0 +1,246 @@
...
+  	$("#preview-chart").highcharts('StockChart',chartOptions);
+  }
+
+  function updateChart(){
+  	Highcharts.setOptions(themeConf);
+  	$("#preview-chart").highcharts().destroy();
+  	createChart();
+  }
+
+  function openConfirmDialog(){
```
arnabk: @4p00rv This entire thing is duplicated from previous theme code. Can you think of reusing the code?

arnabk:
@4p00rv 

- Start with some defaults. I see all colors as white when it starts. If user used custom theme earlier, then start with that. If user started first time or they did not use custom theme prior to opening this dialog, then show some default combination. white color for all looks odd

- Do not start with line chart. Start with candle chart. I should be able to change the candle combination. Check the screenshot attached in the task. Its taken from MT4 platform

- We do not need indicator color because indicator colors are selected when indicators are loaded on chart

- I am not sure if you could easily change Tooltip background for our charts. Make sure that it works and you are not going to break our CrossHair chart options menu item while trying to implement this

- Refer to my comments in Github. I tried to add line by line comments where needed. 

- Why have you not used rivetjs for this? This is a perfect candidate for rivetjs. Code will be much cleaner

src/themes/custom_theme/custom_theme.js
4p00rv: @arnabk I'm not sure how to achieve that.

src/themes/custom_theme/custom_theme.js
4p00rv: Yes @arnabk  this is necessary because once the highcharts options are set for one chart it automatically applies to all new chart which will be created after execution of the Highcharts.setOptions() statement.

src/themes/custom_theme/custom_theme.js
4p00rv: Removed this. Also, the code is from charts.js

src/themes/custom_theme/custom_theme.css
```diff
@@ -0,0 +1,18 @@
+.custom_theme {
+	display: inline-block;
+}
+.custom_theme input {
+    width: 100px;
+}
+#properties {
```
arnabk: @4p00rv Use .custom_theme #properties 
Its better to use .properties than #properties because properties is very common name and could be used anybody in other places of webtrader

src/themes/custom_theme/custom_theme.css
```diff
@@ -0,0 +1,18 @@
...
+	float: right;
+	width: 250px;
+	margin-left: 10px;
+	max-height: 400px;
+	overflow: auto;
+}
+#preview-chart {
```
arnabk: @4p00rv Same comment here

src/themes/custom_theme/custom_theme.js
arnabk: @4p00rv Please review how download.js is coded. In main.js we register the call once.

src/themes/custom_theme/custom_theme.js
arnabk: @4p00rv Yes, agreed. That's a valid point I forgot

src/themes/custom_theme/custom_theme.js
arnabk: @4p00rv We don't need this, so removing the code is fine. 
But for your info, when you copied from charts.js, you missed to copy the click interceptor.

```
chart: {
                    events: {
                        load: function () {
this.credits.element.onclick = function() {
                                window.open(
                                    'http://www.binary.com',
                                    '_blank'
                                );
                            }
            }
}
```

src/themes/custom_theme/custom_theme.css
4p00rv: @arnabk sure, noted.

4p00rv:
@arnabk fixed the issue that you pointed out.

src/themes/custom_theme/custom_theme.html
```diff
@@ -0,0 +1,74 @@
+<div class="custom_theme" title="Customize appearance of chart">
+	<div id="properties">
+			<table width="100%">
+				<tr>
+					<td><strong>Chart background</strong></td>
+					<td>
+					<input value="" id="temp_box" type="text"/>
```
4p00rv: @arnabk had to add this ("not visible") text box in order to remove focus off colorpicker. Tried using `blur` function but it didn't work because the colorpicker popup can't be closed using that.

arnabk:
@4p00rv I after second review, these are the comments - 

- Should require when needed (If you want me to explain it further, lets chat on this)
- Now it shows Highcharts.com as link (This is straight forward. Just copy the code from charts.js)

PR #415, opened 2016-06-22
fixed purchase button

4p00rv:
This PR fixes purchase button for firefox.


arnabk:
@4p00rv Have you tested this in FF, Opera, Chrome, Edge? You can use our browserstack account

PR #433, opened 2016-07-02
Change candle color to green-red for affiliate page

4p00rv:


arnabk:
@4p00rv Any affiliates related change needs to go to affilaites/affiliates.js file. I have copied your code over there. Also do not use shadow : true because it has caused performance issues.

PR #437, opened 2016-07-02
Proposal for client side translation.

aminroosta:
@arnabk  Here is my proposal on how to handle translation, In webtrader we always inject strings to the page, even html files are loaded as strings and then converted to DOM objects using jquery. I've avoided touching DOM elements this time and instead added an `i18n` function to strings to translate them.

Profit table page is updated using this approach to demonstrate this proposal.


arnabk:
@aminroosta 
Everything looks pretty good. Few comments - 
- Validate the lang parameter. We have to allow only supported lang parameters. If unsupported parameter is passed, we need to fall back to EN
- How about moving the definition of  String.prototype.i18n to utils.js?
- i18n folder should not be in src. In fact this folder should be created dynamically by the grunt build process
- Create po folder in the root folder where all our PO files should be. These files will be translated by our translators and later grunt build will compile them into dist/**/i18n folder, where necessary

PR #463, opened 2016-07-20
Added new images

4p00rv:


arnabk:
@4p00rv Can you delete the SVG files not being used anymore? Just do a search in the src folder just to make sure that it is not being used in other places.

4p00rv:
@arnabk Images aren't used anywhere. I have removed those files.

PR #467, opened 2016-07-26
Redesign timepicker and move code to tradeDialog.css

4p00rv:


arnabk:
@4p00rv Why did you move the code to tradeDialog.css?

4p00rv:
The code for timepicker was earlier in tradeDialog.css. So, I thought that the code for this should also be in tradeDialog. I will move it back to main.css

4p00rv:
@arnabk made the changes.

PR #468, opened 2016-07-26
fixed current overlay list

4p00rv:


arnabk:
@4p00rv Something is broken in this fork. When I tried to add overlay instrument, it does not work. Moving that refresh call above might have other side effects.
Also https://github.com/binary-com/webtrader/pull/465 has some problem. You cannot check for isInstrument because isInstrument will also be true for the main series. You have to discard series[0] from that calculation since it series[0] will always be the main series

PR #470, opened 2016-07-28
fix overlay count

4p00rv:


src/charts/overlay/overlayManagement.js
```diff
@@ -104,9 +104,9 @@ define(['websockets/binary_websockets', 'common/rivetsExtra' , 'lodash'], functi
                   charts.overlay(containerIDWithHash, symbol, displaySymbol, delay_amount)
                         .then(function() {
                           var overlay = { symbol: symbol, displaySymbol: displaySymbol, delay_amount: delay_amount};
-                          dialog.trigger('chart-overlay-add', overlay);
                           //Waiting for overlays to be applied.
                           setTimeout(function(){
```
arnabk: @4p00rv In future, you can use _.defer() instead of setTimeout()

4p00rv:
@arnabk Noted your comment, and the made the change. Thanks!

PR #473, opened 2016-07-29
use binary style table

4p00rv:


arnabk:
@4p00rv are these values copied from binary style? Is there a way to use directly from binary style and not hard code it here since binary style might change later?

4p00rv:
@arnabk Yes, these values are taken from binary-style. The binary.css is not able to override the existing dataTable-css, so, I had to hardcode it.

arnabk:
@4p00rv We have to come up with some idea where in datatable styles are loaded first and then binary style. In that way, binary style will be able to override.

4p00rv:
@arnabk what you propose is very unlikely to work because the specificity of data-table styles is higher than that of binary style. One solution that I can think of is that we remove data-tables style. I'am not sure how it would affect other tables though.

4p00rv:
@arnabk please review now, I have removed most of the changes made to main.css, kept it minimalistic.

PR #498, opened 2016-08-17
Fix typo

4p00rv:


arnabk:
I believe this will require changes in translation file. Do not modify translation file now. It creates conflicts. I will see what I can do.

4p00rv:
Ok👍

4p00rv:
@arnabk I have made the changes. Let me know if this is not what you are looking for.

PR #501, opened 2016-08-18
Fix for trade issue

aminroosta:


.tern-project
```diff
@@ -0,0 +1,16 @@
+{
```
arnabk: @aminroosta Whats this file for?

arnabk:
Approved

.tern-project
aminroosta: [ternjs](http://ternjs.net/) enables autocomplete and code analyzes for javascript files.
Many editors have plugins that work with it, this file is the project configuration for ternjs.

arnabk: @aminroosta This is bit personalized. I use webstorm and it works perfectly fine. It has superior auto complete feature and unfortunately, it is not free. please remove this. We dont need it. For code analysis, we should use eslint. We are already ES6, we should go the ESLINT rout

aminroosta: @arnabk Done :-)

PR #513, opened 2016-08-20
Prevent JPY accounts from loging into webtrader

aminroosta:


arnabk:
@aminroosta Can you please submit this PR again to master? This needs to be released in master

PR #524, opened 2016-08-24
Fixed asian chart

4p00rv:


src/trade/tradeConf.js
```diff
@@ -240,7 +240,7 @@ define(['lodash', 'jquery', 'moment', 'websockets/binary_websockets', 'common/ri
               if(this.category === 'Asians') {
                 //https://trello.com/c/ticslmb4/518-incorrect-decimal-points-for-asian-average
                 var avg = this.average().toFixed(decimal_digits + 1);
-                return {value: avg, label:'Average ('.i18n() + avg + ')', id: 'plot-barrier-y'};
+                return {value: avg*1, label:'Average ('.i18n() + avg + ')', id: 'plot-barrier-y'};
```
4p00rv: @arnabk Although I have fixed this but I don't understand why do we need to multiply the value with 1. Can you help?

4p00rv:
@arnabk #523  also fixes the same issue. Please merge any one of these.

src/trade/tradeConf.js
binarytom: `.toFixed()` returns a string, the *1 converts it back into a number. If you want to make the behaviour clearer, you could use `parseFloat(avg)` instead.

4p00rv: @binarytom thanks for the explanation. It makes sense now :+1:

PR #526, opened 2016-08-24
fix script errors

4p00rv:
Deployed [here](http://4p00rv.github.io/webtrader/fix_script_errors).

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -77,6 +77,16 @@ requirejs.config({
     }
 });
 
+requirejs.onError = function (err) {
+    //Avoiding script errors on timeout. Showing a warning so that developers can track wrong path errors on local servers.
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -77,6 +77,16 @@ requirejs.config({
...
 });
 
+requirejs.onError = function (err) {
+    //Avoiding script errors on timeout. Showing a warning so that developers can track wrong path errors on local servers.
+    if (err.requireType === 'scripterror') {
+        console.warn(err); 
+        return;
+    }
+
+    throw err;
```
arnabk: @4p00rv Is this not going to trigger TrackJS logging?

4p00rv: Warnigs are not tracked, I confirmed it.

PR #527, opened 2016-08-25
Share button hist data

4p00rv:
Deployed [here](https://4p00rv.github.io/webtrader/share_button_hist_Data).


src/download/download.css
```diff
@@ -1,55 +0,0 @@
-.download_window {
```
arnabk: @4p00rv You deleted this file. Have you removed its reference?

src/download/download.scss
```diff
@@ -0,0 +1,141 @@
+.download_window {
+    background-color: white !important;
```
4p00rv: @arnabk it is here.

PR #542, opened 2016-08-30
avoid reality check for virtual accounts

4p00rv:


src/websockets/binary_websockets.js
```diff
@@ -358,7 +358,7 @@ define(['jquery', 'text!oauth/app_id.json', 'common/util'], function ($, app_ids
           }
           var oauth = local_storage.get('oauth');
           if(!oauth) {
-            return promise.reject({ message: 'Account token not found.'.i18n() });
```
4p00rv: promise is undefined here.

PR #561, opened 2016-09-08
Update translations

4p00rv:


arnabk:
@4p00rv When you want to change translation files, please let me know first so that I can lock the translation branch and then you can do the modification. Otherwise it always conflicts because the translators are continuously translating something which gets pushed by weblate periodically.

For now, let me know conflicts are resolved here

4p00rv:
@arnabk Resolved the conflicts.

arnabk:
@4p00rv Let me know when you are online. We have to do it together, else it will be in conflict always.

PR #565, opened 2016-09-09
Use currency in tables

4p00rv:


arnabk:
@4p00rv Trade dialogs also show currency at the bottom(account balance). Check with @aminroosta to get more info.

4p00rv:
@arnabk if you mean the following it is done. Trad menu

PR #596, opened 2016-10-17
close chart menus on click

4p00rv:


src/charts/chartOptions.js
```diff
@@ -367,6 +366,16 @@ define(['jquery', 'common/rivetsExtra', "charts/chartWindow", "charts/charts", '
                     });
                 }
 
+                // Add event only once.
+                !isListenerAdded && $('body').on('click', function(event){
```
4p00rv: I tried that but it didnt work because of event propagation. The propagating event caused the _one_ event to execute right after it was added.

PR #601, opened 2016-10-20
Increase chart height

4p00rv:


src/common/util.js
```diff
@@ -26,6 +26,13 @@ function isLineDotType(type) {
     return type === 'linedot';
 }
 
+function isAffiliates(){
```
4p00rv: @arnabk Sure, I will create another PR for those changes.

PR #615, opened 2016-11-11
White labelling

4p00rv:


src/charts/chartOptions.html
```diff
@@ -119,7 +119,7 @@
             <div class="cell" rv-show="showShare">
                 <div class="chartOptions_button" rv-on-click="toggleExportSelector" rv-class-button_selected="showExportSelector">
                     <span class = "shareButton" data-balloon="Share chart" data-balloon-pos="down">
-                        <div class="img"><img src="images/share.svg"></div>
+                        <div class="img"><img src='data:image/png;base64,R0lGODlhFAAUAIAAAP///wAAACH5BAEAAAAALAAAAAAUABQAAAIRhI+py+0Po5y02ouz3rz7rxUAOw==' class="share-icon"></div>
```
arnabk: @4p00rv You added image in the HTML file. It will not go through out image optimization grunt command.

4p00rv:
@arnabk actually it doesn't need to go through image optimization, the src
is just a mere placeholder to overcome weird border problem and is just a
few bytes.

src/charts/chartOptions.html
arnabk: Is it deployed anywhere? I would like to have a look at it

arnabk:
@4p00rv I reverted https://github.com/binary-com/webtrader/pull/614
Please include that change here

Gruntfile.js
```diff
@@ -160,6 +160,14 @@ module.exports = function (grunt) {
                     from: '<description>',
                     to: "<%=pkg.description%>"
                 }]
+            },
+            style : {
```
arnabk: @4p00rv There is a `replace` task. Do you think that it would be good if we move this inside that? Just create two different tasks inside it. One for the previous one and one for this new thing you have done here. We have been doing this for all other cases. You can check copy, server, gh-pages etc as an example

arnabk:
@4p00rv 
Asset Dialog input text issue http://prntscr.com/d7usn9
Font issue with the language menu http://prntscr.com/d7usyb

src/navigation/navigation.html
```diff
@@ -9,13 +9,13 @@
...
+                <div class="logo">
+                    <img class="responsive" src="<style-url>/images/logo/symbol.svg">
                 </div>
-                <div>
-                    <img class="responsive" src="https://www.binary.com/images/pages/binary-type-logo.svg" style="width:80%;">
+                <div class="binary-logo-text">
+                    <img class="responsive" src="<style-url>/images/logo/type.svg" style="width:80%;">
                 </div>
             </a>
         </div>
```
negar-binary: @4p00rv can you also use this logo code in 404.html and any other place that is using it?

PR #657, opened 2016-11-25
Switch to VRTC on selfExclude

4p00rv:


arnabk:
@4p00rv Here are comments from Amin.


amin	[10:15 AM]  
I think apoorv can use `switch_account: function(id) { ...` function defined in the `binary_websockets.js` file.

[10:17]  
He also should use ` var invalidate = function(){ ...` instead of `trigger("logout")` to properly logout the user (at least what i think when i look at the pr changes).

Please chat with Amin on this to get more information.

4p00rv:
@arnabk I didn't use either of the functions because we're using is_authenticated check before proceeding with the other statement. is_authenticate is false because the authentication was never successfull

4p00rv:
@arnabk @aminroosta I have used `switch_account` & `invalidate`, but in order to use them I had to remove the `is_authenticated` check. I went through the code and don't think removing `is_authenticated` check would cause any issue.

aminroosta:
>  I have used switch_account & invalidate, but in order to use them I had to remove the is_authenticated check.

@4p00rv 
Just looking at diff, it seems to me that when you call `invalidate()` and `switch_account()` the session is already authenticated (`is_authenticated == true`), if my guess is correct then there should not be any problem leaving that check.

> I went through the code and don't think removing is_authenticated check would cause any issue.

I too went through the code and it seems this check will not happen because other dialogs are well guarded not to call these methods unless the user is already authenticated, however the purpose of this check is to avoid issues such as `null reference object` ... keep it if you can :-)

4p00rv:
@aminroosta I too thought
> that when you call invalidate() and switch_account() the session is already authenticated.

But it is not, since the call to `authenticate` was never succesful because it returned an error and `is_authenticate` was never set to true.

aminroosta:
@4p00rv Seems good to me then 👍

PR #659, opened 2016-11-28
Opera unsupported

4p00rv:
jQuery is not supported in Opera v11. Moved the check for compatibility outside the jquery block to successfully check for compatible browsers.


4p00rv:
[Here's](https://bugs.jquery.com/ticket/14032) a bug filed for the same issue.

PR #660, opened 2016-12-01
Template error

4p00rv:


src/common/util.js
```diff
@@ -333,6 +333,32 @@ if (!Array.prototype.includes) {
     return false;
   };
 }
+
+if (!Array.prototype.find) {
```
arnabk: @4p00rv Where are you using this? Can we not use _.find() ?

4p00rv:
@arnabk Thanks it never hit me that I can use lodash. Now I have used it, and it is neat! :smile:

PR #663, opened 2016-12-07
Auto logout user

4p00rv:
@arnabk I commented the code which was used for resubscribing when the websocket connection is closed, maybe in future if we decide to recert back to not logging out the user on connection close we can just un-comment this code.


4p00rv:
@arnabk should I close the PR?

PR #668, opened 2016-12-14
Use binary-style for date-picker

4p00rv:


src/windows/windows.js
```diff
@@ -301,7 +301,7 @@ define(['jquery', 'lodash', 'navigation/navigation', 'windows/tracker', 'jquery.
 
             var options = {
                 showOn: 'both',
-                numberOfMonths: 2,
```
4p00rv: @arnabk Do we really need 2 months?

arnabk: @4p00rv Not really. Is it causing any issues?

4p00rv: @arnabk the issue was that the binary-style is not supplying a css-style for the other month, this caused the date picker to look funny. I have checked, changing the month to 1 won't cause any issues for webtrader.

PR #685, opened 2017-01-11
bring in changes from master

arnabk:


src/instruments/instruments.js
```diff
@@ -75,18 +75,19 @@ define(["jquery", "jquery-ui", "websockets/binary_websockets", "navigation/menu"
...
-                    .then(function (data) {
-                        chartable_markets = menu.extractChartableMarkets(data);
-                        refresh_active_symbols();
-                        require(['websockets/binary_websockets'],function(liveapi) {
-                          liveapi.events.on('login', refresh_active_symbols);
-                          liveapi.events.on('logout', refresh_active_symbols);
-                        });
-                        /* refresh menu on mouse leave */
-                        var instruments = $("#nav-menu").find(".instruments").on('mouseleave', refresh_active_symbols);
+            return liveapi
```
arnabk: @aminroosta This line was missed out and affiliates.js was dependent on this. It was expecting Promise return.

src/instruments/instruments.js
```diff
@@ -75,18 +75,19 @@ define(["jquery", "jquery-ui", "websockets/binary_websockets", "navigation/menu"
...
+                .then(function (data) {
+                    chartable_markets = menu.extractChartableMarkets(data);
+                    refresh_active_symbols();
+                    require(['websockets/binary_websockets'],function(liveapi) {
+                      liveapi.events.on('login', refresh_active_symbols);
+                      liveapi.events.on('logout', refresh_active_symbols);
                     });
+                    /* refresh menu on mouse leave */
+                    var instruments = $("#nav-menu").find(".instruments").on('mouseleave', refresh_active_symbols);
+                    return chartable_markets;
```
arnabk: @aminroosta affiliates.js is expecting Promise

arnabk: @aminroosta When a user is logged in binary.com, does it mean that they are also automatically logged into webtrader? If true, can we make sure that this promise waits until refresh_active_symbols function is ready so that affiliates.js also uses the same valid list for charting?

PR #688, opened 2017-01-15
reconnect websocket after connection error

aminroosta:


src/websockets/binary_websockets.js
```diff
@@ -36,8 +36,8 @@ define(['jquery', 'text!oauth/app_id.json', 'common/util'], function ($, app_ids
         ws.addEventListener('message', onmessage);
 
         ws.addEventListener('error', function(event) {
-            $.growl.error({message: 'Connection error. Refresh the page.'.i18n().i18n()});
-            //Clear everything. No more changes on chart. Refresh of page is needed!
+            $.growl.error({message: 'Connection error.'.i18n()});
```
arnabk: @aminroosta We don't have translation for this string. Could you please also run extract.py and submit the translation file changes too? I will get it translated.

PR #692, opened 2017-01-17
Fix NaN in table view

4p00rv:


src/websockets/ohlc_handler.js
```diff
@@ -30,6 +30,11 @@ define(['websockets/binary_websockets','charts/chartingRequestMap','jquery','jqu
     liveapi.events.on('candles', function (data) {
         var key = (data.echo_req.ticks_history + data.echo_req.granularity).toUpperCase();
         for ( var index in data.candles ) {
+            //This is for microsoft products. 
+            //They some how also include the include() while iterating over the object.
+            if(index === 'includes'){
```
arnabk: @4p00rv I am not sure if I like this, but can we do it in a better way? That is just do `(data.candles || []).forEach()`

My recommendation is we should use ES6/ES7 wherever possible so that we are improving code gradually.

4p00rv: Sure, I will go ahead and put `forEach`

4p00rv:
@arnabk Please re-review

src/websockets/ohlc_handler.js
```diff
@@ -29,34 +29,24 @@ define(['websockets/binary_websockets','charts/chartingRequestMap','jquery','jqu
...
-            if(index === 'includes'){
-              return;
-            }
-            var eachData = data.candles[index],
-                    open  = parseFloat(eachData.open),
-                    high  = parseFloat(eachData.high),
-                    low   = parseFloat(eachData.low),
-                    close = parseFloat(eachData.close),
-                    time  = parseInt(eachData.epoch) * 1000;
+        data.candles.forEach(function(eachData){
```
arnabk: @4p00rv Are you sure data.candles or data.history.times are not going to be undefined/null? What about doing it (data.candles || [])

Also have you tested these changes if it is working in all browsers

4p00rv: Yes I have tested it on all the major browsers including IE and Edge.

For the second one there is one possibility that the returned array could be empty, which `forEach` handles pretty well.

PR #704, opened 2017-01-20
Set default currency based on account type

4p00rv:


src/cashier/currency.es6
```diff
@@ -4,6 +4,7 @@ import windows from 'windows/windows';
 import rv from 'common/rivetsExtra';
 import html from 'text!cashier/currency.html';
 
+require(['common/util'])
 require(['text!cashier/currency.html']);
```
arnabk: @4p00rv You already imported html. Do you still have to do this?

src/common/util.js
```diff
@@ -401,7 +401,7 @@ var Cookies = {
       return loginids.map(function(_id) { return _id.id }).indexOf(id.id) === -1;
     });
 
-    return oauth_loginids ? oauth_loginids : loginids
+    return oauth_loginids.length > 0 ? oauth_loginids : loginids
```
arnabk: @4p00rv Do this
`oauth_loginids && oauth_loginids.length > 0 ? oauth_loginids : loginids`

src/cashier/currency.es6
4p00rv: @arnabk I think there's no need to import the html, although it was already there before I made the change, so I thought of not removing it.

src/common/util.js
4p00rv: @arnabk The statement above, which initializes the `oauth_loginids`, atleast returns an empty array, that's why I didn't put the extra check.

arnabk: @4p00rv We do not have unit tests and I am worried that when you make that statement, are you really considering all scenarios where it can never be null or undefined

src/cashier/currency.es6
arnabk: @4p00rv Why not we delete that require() statement and keep that import and see if it works. I would prefer to use the import rather than require to reduce one network call here

src/common/util.js
4p00rv: @arnabk Yes, I did. Although I think I should add the extra check, in order to be certain that it never breaks, even if someone changes the statement above.

4p00rv:
@arnabk I have updated the checks.

arnabk:
@4p00rv Thanks. I have one more comment above. Could you please try that and let me know? If that works, I would like to do the same thing in some other places.

4p00rv:
@arnabk Please review this. Also I removed the extra require statement and it seem to work fine.

src/cashier/currency.es6
```diff
@@ -4,7 +4,7 @@ import windows from 'windows/windows';
 import rv from 'common/rivetsExtra';
 import html from 'text!cashier/currency.html';
 
-require(['text!cashier/currency.html']);
+require(['common/util']);
```
arnabk: @4p00rv How about doing import 'common/util'?

4p00rv: That seems to work too, although I couldn't find any similar statement in the documentation.

PR #707, opened 2017-01-23
fix multiple reality-check pop ups

4p00rv:


src/realitycheck/realitycheck.js
```diff
@@ -98,6 +96,11 @@ define(["jquery", "windows/windows", "websockets/binary_websockets", "lodash", '
 
                                 require(['text!realitycheck/realitycheck.html', "css!realitycheck/realitycheck.css"], function(html) {
                                     var div = $(html).i18n();
+                                    // Is there a better option?
+                                    if(win){
+                                        resolve();
+                                        return;
+                                    }
                                     win = windows.createBlankWindow($('<div/>'), {
```
4p00rv: @arnabk The problem here's we're calling the init() twice. The second call is so close to previous one that it gets through the `if(win)` check to next statement. The best way I found to fix the problem was to move the check inside the promise. Do you know a better approach to this?

arnabk: @aminroosta The best way to implement this is to use combination of promise and event manager. You can get example from windows.js/binary_sockets.js how @aminroosta did it. What you can do is, wherever init() is being called from, return a promise(which we are already doing). Then resolve all these promises when "ready" event is emitted.

4p00rv: @arnabk I will try that out. Thanks!

4p00rv:
@arnabk  I made some changes based on your previous comment. Please review.

src/realitycheck/realitycheck.js
```diff
@@ -162,11 +162,15 @@ define(["jquery", "windows/windows", "websockets/binary_websockets", "lodash", '
 
     var oauthLogin = function() {
         logout();
-        if(!local_storage.get("authorize").is_virtual)
-            init().then(function() {
+        if(!local_storage.get("authorize").is_virtual){
+            var p = init();
+            promises.push(p);
```
arnabk: @4p00rv When promises are resolved, are you clearing the array `promises` somewhere? you do not want these promises' resolve to be called later if someone calls init() again when win is already loaded.

4p00rv: @arnabk Yes, that would be correct approach. I have made the change.

PR #709, opened 2017-01-23
Show balance in portfolio the first time

aminroosta:


src/portfolio/portfolio.es6
```diff
@@ -0,0 +1,306 @@
...
+    transaction = _.last(transaction);
+    $target.addClass('button-disabled');
+    require(['viewtransaction/viewTransaction'], function(viewTransaction){
+    viewTransaction.init(transaction.contract_id, transaction.transaction_id)
+                    .then(function(){ $target.removeClass('button-disabled'); });
+    });
+}
+
+function initPortfolioWin() {
+    require(['css!portfolio/portfolio.css']);
```
arnabk: @aminroosta Since you have refactored this code and moved it to ES6, could you please do import statement instead of require. This way it would save one network call.

PR #717, opened 2017-01-27
Help repository

4p00rv:


src/help/help.es6
```diff
@@ -0,0 +1,305 @@
...
+		if(query.length > 0){
+			state.current.sublist = sublist_items.filter((item) => {
+				return item.text.toLowerCase().indexOf(query) != -1;
+			});
+			state.current.sublist && state.current.sublist.length && state.getContent(state.current.sublist[0].url);
+		}
+	}
+
+	//Concat all the sublist items into one array so that we can later use it for searching.
+	for (var key in state.sublist){
```
arnabk: @4p00rv In ES6, please do not use `var`. `var` has got some other side effects. Use `const` or `let` as per situation

arnabk:
@4p00rv Since I merged our translation, this branch has got some conflicts. Please fix them.

arnabk:
@4p00rv I tried to test and this is what I see.
screen shot 2017-01-28 at 11 46 48 pm

I also wanted to do one more changes to this. (And you can submit a separate PR for it, if dont want to include in this). I wanted to show "?" in all indicator dialogs which when clicked would take them directly to the specific section of the help window. This icon show not show up when access using affiliates=true

arnabk:
@4p00rv One issue is still not resolved.
screen shot 2017-02-01 at 9 58 31 am

src/cashier/withdraw.es6
```diff
@@ -7,12 +7,11 @@ import liveapi from 'websockets/binary_websockets';
 import windows from 'windows/windows';
 import rv from 'common/rivetsExtra';
 import currencyDialog from 'cashier/currency';
-import _ from 'lodash';
+import {debounce} from 'lodash'
```
4p00rv: @arnabk It turns out that the syntax for loading a specific function from lodash is as described above. `import debounce from 'lodash/debounce'` doesn't work.

arnabk: @4p00rv doing `import _ from 'lodash'` and `import {debounce} from 'lodash'` would result in same size file. If you do `import debounce from 'lodash/debounce'`, that would save lot of file size. I have used that in many projects. You can also check the node_modules/lodash folder and find out more about it.

4p00rv: @arnabk Actually `import {debounce} from 'lodash'` imports only the debounce function from lodash.

arnabk: @4p00rv I read an article long time back on this. I am not sure what and where you are referring which says that this is the right way. But the truth is, its not the right way. It will make the file bulky. You can check the lodash code. The way to test it out is to check the generated file size with two approaches.

I could not find the original article where I read, but a quick google got me this.
http://stackoverflow.com/questions/35250500/correct-way-to-import-lodash

Read it. It will be useful. If you ever get into React/Redux development, its kind of *mandatory* to use `import debounce from 'lodash/debounce'` coding, considered under best practices.

4p00rv: @arnabk I tried that and it doesn't seem to work because on `import debounce from 'lodash/debounce'` it actually sends a request to **https://localhost/<version>/lib/lodash/dist/lodash.min/debounce.js** which doesn't exist. We might have to change something in the config to make it work.

arnabk: @4p00rv I figured out why it is sending separate requests for imported module. @aminroosta  is using `transform-es2015-modules-amd`. I am not sure if I would like to continue using this, but changing this would require some effort. For now I will let it go and when we move completely to webpack, we can get rid of this AMD based babel plugin. webpack supports dividing code into `chunks`.

PR #723, opened 2017-01-31
Fix withdraw issues

4p00rv:


4p00rv:
@arnabk We had some changes in the backend which now requires `occupation` field for `new_account_maltainvest` call.

src/cashier/withdraw.es6
```diff
@@ -85,10 +85,19 @@ class Withdraw {
       route: { value: 'menu'},
       empty_fields: {
         validate: false,
-        clear: _.debounce(() => ( state.empty_fields.validate = false ), 4000),
```
arnabk: @4p00rv I am merging this for now, but could you please just import the function that we really need here. In ES6 import style, we can do `import debounce from 'lodash/debounce'`
This will only import the module we need and later when withdraw.es6 is converted to ES5, it will be smaller in size

PR #776, opened 2017-02-17
Crs tax information

4p00rv:


Gruntfile.js
```diff
@@ -53,7 +53,8 @@ module.exports = function (grunt) {
                             '!binary-com-jquery-dialogextended/**', 'binary-com-jquery-dialogextended/jquery.dialogextend.min.js',
                             'binary-com-jquery-ui-timepicker/jquery.ui.timepicker.js', 'binary-com-jquery-ui-timepicker/jquery.ui.timepicker.css',
                             'text/text.js',
-                            'regenerator-runtime/*'
+                            'regenerator-runtime/*',
+                            'chosen-js/*'
```
arnabk: @4p00rv I don't see this dependency in package.json. Where is it? And any specific reason why we need this and cannot use Jquery-UI selectbox?

4p00rv:
@arnabk I thought yarn add "library" is suppose to add it to yarn.lock file. Apparently change to yarn.lock are not here.

Gruntfile.js
4p00rv: @arnabk since we allow multiple values for `tax_residence` value, this library is to support multiselect feature for it.
PS: Apparently I missed out on adding the changes to git for the library. it is added now.

arnabk:
@4p00rv Yes, if you do `yarn add <library>`, it will add to yarn.lock as well as to package.json

PR #800, opened 2017-02-22
Fix chart layout

4p00rv:


arnabk:
@4p00rv This URL does not work http://4p00rv.github.io/webtrader/chartWindow-to-es6

4p00rv:
@arnabk I'm sorry I didn't deploy it.  Should be up in a minute.

PR #801, opened 2017-02-22
Crs_tax_info

4p00rv:


arnabk:
@4p00rv Let's take the safe route. I will merge development to master otherwise I will have to review each changes to make sure that they are matching with the original one.

4p00rv:
@arnabk I guess that is a better approach.

PR #809, opened 2017-02-23
fix css issue

4p00rv:


src/demo/affiliates_demo_iframe.html
```diff
@@ -45,7 +45,7 @@
     <script charset="utf-8">
         $(document).ready(function() {
 
-          var ws = new WebSocket("wss://ws.binaryws.com/websockets/v3");
+          var ws = new WebSocket("wss://ws.binaryws.com/websockets/v3?app_id=0");
```
arnabk: @4p00rv Use correct app_id. Use the app_id that we are using for production. Should be in app_id.json

src/demo/affiliates_demo_iframe.html
```diff
@@ -54,6 +54,7 @@
           };
           ws.onmessage = function(event) {
             var data = JSON.parse( event.data );
+            console.log(data)
```
arnabk: @4p00rv Remove

4p00rv:
@arnabk Done.

PR #815, opened 2017-02-24
use exit spot & time instead of sell spot & time

4p00rv:


arnabk:
@4p00rv Please submit this to master

PR #837, opened 2017-03-07
Fix view transaction

4p00rv:


arnabk:
@4p00rv Its difficult to find out how many changes you really did and how many are just formatting. I see lot of spaces added and there by hard for me to review the real diff.

4p00rv:
@arnabk I have changed the file to contain only the changes that I introduced.

arnabk:
@aminroosta These are bit serious changes. Could you please double confirm that it is not going to break our trade module? I know that we stabilized the trade module after lot of back and forth between clients and us.

src/trade/tradeConf.es6
```diff
@@ -147,26 +137,16 @@ const register_ticks = (state, extra) => {
...
 
    fn = liveapi.events.on('proposal_open_contract', (data) => {
       const contract = data.proposal_open_contract;
       if(contract.contract_id !== extra.contract_id) return;
       entry = contract.entry_tick_time * 1;
       expiry = contract.date_expiry * 1; /* date_expiry gets updated when contract is settled !!! */
-      if(!tracking_timeout_set)
-         track_ticks();
+      if (contract.current_spot_time >= entry)
+         add_tick(data.proposal_open_contract);
```
aminroosta: @arnabk @4p00rv By the time `porposal_open_contract` event is registered there might already be some ticks that have been passed. In other words not all ticks for a contract are returned with this api.

The track_ticks() function and the last_1000_ticks[] array that are now removed are intended to fix this issue.
@arnabk With those stuff gone the trade module will be broken.

src/trade/tradeConf.es6
```diff
@@ -220,15 +200,6 @@ export const init = (data, extra, show_callback, hide_callback) => {
...
          array: [],
-         average: () => {
-            const ticks = state.ticks;
-            const array = ticks.array;
-            let sum = 0;
-            for(let i = 0; i < array.length; ++i)
-               sum += array[i].quote*1;
-            const avg = sum / (array.length || 1);
-            return avg;
-         },
```
aminroosta: @4p00rv My guess is that "tick.barrier" will return the new barrier value for Asians. Make sure that's the case.

4p00rv: @aminroosta I have verified that the barrier works fine for asians.

src/trade/tradeConf.es6
4p00rv: @aminroosta @arnabk I would like to disagree. Please refer to the following screenshot: 

some

What I'm doing is similar to what was done earlier, i.e; instead of subscribing to the `on('ticks')` event I'am subscribing to the `proposal_open_contract`. In none of the cases we're sending an api request from **tradeConf.es6**, hence making the delay same for both the requests. Also, I would say opting for `proposal_open_contract` is better because on successful buy we first get a _stray tick_ confirming that the contract was bought successfully and then it starts streaming, giving us ample of time to register the `proposal_open_contract` event

aminroosta:
screen shot 2017-03-09 at 2 30 13 pm

@4p00rv 
> By the time porposal_open_contract event is registered there might already be some ticks that have been passed.

For example in the above example `tick 1` is missed.

aminroosta:
screen shot 2017-03-09 at 2 44 01 pm

@4p00rv Actually i now get your point. but you still can miss a tick.
My previous example was not quite correct, i've added another example.
By the time `liveapi.events.on('propsoal_open_contract', ...)` is registered, there might be some ticks passed.

4p00rv:
@aminroosta As I mentioned earlier that there will be 1 stray response on subscribing to `proposal_open_contract`. I have some screenshots to confirm that.
stray response
first response

As you may notice that the first response doesn't contain `barrier` property in response, so, the backend sends the same tick again and adds `barrier` property to it.

We can still add a fail-safe condition in the code for first tick, which is, if it is the first tick and `current_spot_time` doesn't match with `entry_tick_time` then we first add `entry_tick & entry_tick_time` to array and then add `current_spot & current_time`.

Please let me know if you agree with this approach.

PS: We get all this response in `proposal_open_contract`.

arnabk:
@4p00rv Why do you want to change the trade module code. Believe me, we had real tough situation to get this going. I will be discussing with JY on better backend API for trade module when I am in Malaysia. The existing backend API for trade is not good. Can we not defer any changes to trade module for now? We can focus on something else. I would have merged this changes if we had good test coverage for trade module. We have nothing now for entire Webtrader. That means, whatever exhaustive understanding we might have now, we might still miss some realistic scenario that were coming out when clients started using it. Just do not want those again.

aminroosta:
@4p00rv I did review your changes once again and it seems to me that these changes *might*  actually be ok.
However i want to test it myself, because network delay is hard to reproduce and QA usually are not able to test it properly.

I've been busy with another task, but i will test the next week :-)

aminroosta:
@arnabk I did test @4p00rv changes today and i believe its working fine. 👍

4p00rv:
@arnabk Please do not merge this PR yet, I would like to add a fail safe condition to it.

PR #860, opened 2017-03-29
replace forEach with  `for` in callback functions

4p00rv:
This should make webtrader at least 40-50% faster, the other part is mostly highcharts issues which causes performance issue.

### DIFF
--- a/src/charts/chartingRequestMap.es6
+++ b/src/charts/chartingRequestMap.es6
@@ -37,8 +37,8 @@ export const barsLoaded = function(instrumentCdAndTp) {
     const chartIDList = this[key].chartIDs;
     const processOHLC = this.processOHLC;
 
-    chartIDList.forEach((chartID) => {
-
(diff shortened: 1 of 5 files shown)


arnabk:
@4p00rv @aminroosta Could you please let me know what's the other issue. We can open defect with Highcharts so that they can investigate further. Their requirement is, you have to recreate the issue in JSFiddle so that they could investigate.

@4p00rv 
Also can we change all `forEach` usage in entire codebase?

4p00rv:
@arnabk It is not required to change it everywhere, I have changed it at places which are invoked very frequently. We just need to be careful from now on, and not use `forEach` within functions which are called very frequently.

Regarding the highchart issue, it could be because we have a lot of Wrappers around the original function, but I couldn't really figure out exactly which function was causing the performance issue. It could be because of a lot of dom manipulations. I will try to investigate more about it and try to create POC for the issue.

PR #871, opened 2017-04-06
Restore unique dialogs on connection close

4p00rv:


src/websockets/binary_websockets.es6
```diff
@@ -63,7 +64,10 @@ const onclose = () => {
             socket = connect();
             if(local_storage.get('oauth')) {
               api.cached.authorize().then(
-                 () => tracker.reopen_trade_dialogs(trade_dialogs)
+                 () => {
+                       tracker.reopen_trade_dialogs(trade_dialogs);
+                       setTimeout( _ => tracker.reopen_unique_dialogs(unique_dialogs), 0);
```
arnabk: `_`  is normally used for lodash/underscore. If in future we try to use lodash in this file, this could be misleading. In this line, you can just use `()` instead of `_` if you really do not need any variable name

4p00rv:
@arnabk Done.

PR #872, opened 2017-04-07
Fix trade issue

4p00rv:
@arnabk Earlier implementation of trade dialog was on language dependent values returned from backend, which caused various issues on language change.

 These changes are to make the trade dialog language independent and function properly. I have tested all the related functionalities and there's just one problem, trade templates don't work at all.

@aminroosta since you created the file, can you help me understand the implementation for setting template? Is it dependent on rivets bound object mutation?


arnabk:
@aminroosta Please review this change and let me know if its ready for merge.

4p00rv:
@arnabk I have fixed the trade template issue.

PR #878, opened 2017-04-13
Welcome page

4p00rv:


arnabk:
@4p00rv Do you think that it would be good to move the large CSS into its own file? May be SCSS file?

I see that you removed footer from main webtrader page. I remember vaguely we coded some JS in order to move it down as the dialogs are dragged down. Can we get rid of those code?

4p00rv:
@arnabk I think it is a good idea to move the css to a separate file. 

I will remove the js code as well.

4p00rv:
@arnabk Done.

PR #897, opened 2017-04-24
Add indicator option in view historical

4p00rv:


arnabk:
@4p00rv I think you have to hold on to this since webtrader-charts has been created now, indicators.js file belongs to that project and not webtrader.

You can check with @aminroosta for more details

4p00rv:
@arnabk Sure.

PR #911, opened 2017-04-28
Add testing environment

4p00rv:


.travis.yml
```diff
@@ -8,15 +8,24 @@ node_js:
   - "6"
 notifications:
   email:
-    - arnab@binary.com
+    #- arnab@binary.com
```
arnabk: @4p00rv Is it going to stop sending me notification of failed build?

.travis.yml
```diff
@@ -8,15 +8,24 @@ node_js:
...
+  # give Meteor some time to start
+  - "sleep 5"
 script:
-  - git config --global user.email "arnab@binary.com"
-  - git config --global user.name "Arnab Karmakar"
-  - yarn deploy-gh-pages
+  - yarn test
+  #- git config --global user.email "arnab@binary.com"
+  #- git config --global user.name "Arnab Karmakar"
+  #- yarn deploy-gh-pages
```
arnabk: @4p00rv And how are we going to deploy now?

4p00rv: @arnabk I am just testing one of the solutions I found, please don't merge this yet. I tested browser stack and found it to be working, I will revert my changes to that.

4p00rv:
@arnabk You can merge it now.

4p00rv:
@arnabk I have added tests for Login, Chart and Trade. I believe this would be enough to implement testing environment for beta. 

I do have one question regarding the login token that I'm using for logging in to my account, should we move that as well to travis-env variables?

.travis.yml
```diff
@@ -14,9 +14,10 @@ branches:
   - master
   - development
 script:
-- git config --global user.email "arnab@binary.com"
-- git config --global user.name "Arnab Karmakar"
-- yarn deploy-gh-pages
+  - git config --global user.email "arnab@binary.com"
+  - git config --global user.name "Arnab Karmakar"
+  - yarn deploy-gh-pages
+  - yarn test-browserstack
```
arnabk: @4p00rv Why are we running test after deploy? How much time is it taking to run the tests? Are we not supposed to run the tests after build and before deploy? If tests fail, then deploy should fail too.

README.md
```diff
@@ -94,8 +94,20 @@ Some of these parameters are
...
+## Testing
+To run tests locally:
+```
+yarn test
+```
+To run tests on *BrowserStack*:
+
+- Edit `~/.bash_profile` and append the following lines:
+  ```
+  export BROWSERSTACK_USERNAME="${BrowserStack-Username}"
```
arnabk: @4p00rv Is there any way we could pass these as parameters to yarn command?

arnabk:
@4p00rv `I do have one question regarding the login token that I'm using for logging in to my account, should we move that as well to travis-env variables?`
Please provide more information. you should not be using anything from your account. All variables should be using travis ci config file.

.travis.yml
4p00rv: @arnabk I had some reasons for adding it after the deploy, but now it makes more sense to add it before deploy.

4p00rv:
@arnabk I'm sorry for phrasing that sentence ambiguously, what I mean was that I'm using some API tokens which corresponds to some account on Binary.com, these API tokens are hard coded in the tests, visible on our public repo. Moving these tokens to env variables in travis would help us in hiding it.

arnabk:
@4p00rv I got all the code from your repo, then followed the instructions, then ran yarn test. All tests failed. Here is a video https://drive.google.com/file/d/0B-SvgdYmyOyKNTIwNW9HU0wxNkE/view

1. Please update the version of nightwatch used
2. I also noticed that you have different files for different test suites. For example chart.test.js have all the tests for charts. Instead of putting all tests in one file, could you please divide them into individual file. Each file doing one test only and the do an `import` of that file into chart.test.js. So you can have chart.test.js as main file and all other as sub files with one test per file, which gets included into the main file. Create chart folder and dump all chart related tests in that file

4p00rv:
@arnabk the error is because you didn't run `grunt` before running the tests. Please run `grunt` first and then test.

arnabk:
@4p00rv Update the README test section documentation with following information
```
yarn build-all
yarn test
```
As I understand, yarn will fail, if yarn build-all is ran first. You do not have to add `yarn` command in the list because those who are going to follow README, they will probably run yarn the very first command because it is like doing `npm install` after getting any nodejs project.

Also I think the tests folder structure would make more sense if we restructure them like this
screen shot 2017-05-23 at 6 29 28 pm

Each test file should just do one test. For example, login.test.js has login, portfolio, statement etc all of them together. Its better to separate them in each file. If login is needed for each of them, then create a common module under login folder which only does login and that could be used by all others.

PR #913, opened 2017-05-03
Self-exclusion fix

4p00rv:


src/selfexclusion/selfexclusion.es6
```diff
@@ -116,6 +116,7 @@ const settingsData = {
 
         //validation
         $.each(limits, function(index, value) {
+            scope[index] = scope[index] && '' + scope[index];
```
arnabk: @4p00rv I hope this code works the way you want it to work. But would it be more clear if we do this `scope[index] = scope[index] && ('' + scope[index]);` if that is what you wanted?

4p00rv:
@arnabk I think `toString()` will be more readable. Made the change.

PR #920, opened 2017-05-12
Upgrade binary-longcode

4p00rv:


4p00rv:
@ashkanx the last commit should fix the issue with the windows.

ashkanx:
Thanks.

PR #929, opened 2017-05-29
Add notification banner

4p00rv:


src/websockets/binary_websockets.es6
```diff
@@ -149,14 +152,30 @@ const needs_authentication = (data) => {
    return false;
 };
 
+const website_status = (data) => {
```
arnabk: @4p00rv I don't think belongs to this file. Either we write a separate JS/class file where this notification is handled(we can have it inside navigation folder or we move this code to navigation.es6
binary_websockets.es6 is to purely deal with websockets and dispatching events.

4p00rv: @arnabk there is a flag that is used to stop sending any WebSocket requests and buffer them. Instead of moving it to the navigation folder I made a separate folder to hold the logic for this. 

There are a couple of lines of code which deals with checking the flag `is_website_up` and sending all the buffered requests.

src/websockets/binary_websockets.es6
```diff
@@ -83,7 +84,18 @@ const is_connected = () => (socket && socket.readyState === 1);
 
 
 const onopen = () => {
-   /* send buffered sends */
+   //if website is not up then just subscribe to the website_Status request   
+   if(!is_website_up){
+      socket.send({ website_status: 1, subscribe: 1});
+      return;
+   }
+   restore_buffered_requests()
```
arnabk: @4p00rv Please add a `;`

4p00rv:
@arnabk Done.

arnabk:
@4p00rv Based on our conversation

1. move is_website_up to binary_websockets.es6
2. Then Remove the import from binary_websockets
3. Then import binary_websockets in websitestatus
4. Then subscribe to website status event from websitestatus.es6

In 4th step, most probably, this is what you will do -

```
api.events.on('website_status', (data) => {
    website_status(data);
});
```

4p00rv:
@arnabk I have made the changes based on your suggestions.

src/websitestatus/websitestatus.es6
```diff
@@ -0,0 +1,17 @@
...
+    }
+    else {
+        const $ele = $("#msg-notification");
+        const message = $('<div class="error-msg"/>').append(data.website_status.message || 'Seems like our servers are down, we are working on fixing it.'.i18n());
+        $ele.is(':hidden') && $ele.html(message) && $ele.slideDown(500) && $('.webtrader-dialog').parent().animate({ top: '140px' });
+    }
+};
+
+liveapi.events.on('website_status', website_status);
+liveapi.send({ website_status: 1, subscribe: 1 });
```
arnabk: @4p00rv Move this to binary_websockets.es6
Since everything related to triggering and listening is done in there, it should be there. website_status should just listen to events because it handling the view part(basically responding to what to show when there is notification)

PR #934, opened 2017-06-01
remove affiliates

4p00rv:
@arnabk @aminroosta Some of the references of `isAffiliates` `isHideOverlay` & `isShare` were in chartOptions and charts.es6 file. If not already removed, please remove it from there as well.


src/index.html
```diff
@@ -228,18 +233,35 @@
...
             var url = 'https://www.binary.com/' + (local_storage.get('i18n') || { value: 'en' }).value + '/contact.html'
             var win = window.open(url, '_blank');
             win.focus();
         }
+
+        function addLangChangeListener() {
+            $('.languages #select_language li').each((i, el) => {
+                $(el).click(() => {
+                    var anchor = $(el).find('a')[0];
+                    var lang = anchor && /lang=\w+/.test(anchor.href) && anchor.href.match(/lang=(\w+)/)[1];
```
arnabk: @4p00rv Regex is very powerful, but could be expensive and definitely harder to understand. It is not necessary to use RegEx in places where we could write easy reading code.

PR #941, opened 2017-06-15
Revert "Fix console error"

4p00rv:
Reverts binary-com/webtrader#936

### DIFF
--- a/src/windows/windows.es6
+++ b/src/windows/windows.es6
@@ -401,9 +401,6 @@ export const init = function($parentObj) {
    /* wrap jquery ui dialog('destory') to fire an event */
    var original = $.fn.dialog;
    $.fn.dialog = function(cmd) {
-      if(!this.data('ui-dialog')) {
-         return this;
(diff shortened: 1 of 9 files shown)


src/windows/windows.es6
```diff
@@ -401,9 +401,6 @@ export const init = function($parentObj) {
    /* wrap jquery ui dialog('destory') to fire an event */
    var original = $.fn.dialog;
    $.fn.dialog = function(cmd) {
-      if(!this.data('ui-dialog')) {
-         return this;
-      }
```
aminroosta: @4p00rv Just keep these 3 lines, i do think reverting this PR is the best solution. 👍

4p00rv: @aminroosta Actually now that you pointed out it turns out that the problem is with that line itself. With that line removed everything seems to work fine.

4p00rv:
Closing this since #942 fixes the issue.

PR #1013, opened 2017-08-02
Enable crowdin incontext translation

4p00rv:


src/main.html
```diff
@@ -1,66 +1,71 @@
...
+    <link rel="icon" type="image/png" sizes="160x160" href="<style-url>/images/favicons/favicon-160x160.png">
+    <link rel="icon" type="image/png" sizes="96x96" href="<style-url>/images/favicons/favicon-96x96.png">
+    <link rel="icon" type="image/png" sizes="16x16" href="<style-url>/images/favicons/favicon-16x16.png">
+    <link rel="icon" type="image/png" sizes="32x32" href="<style-url>/images/favicons/favicon-32x32.png">
 
     <link rel="chrome-webstore-item" href="https://chrome.google.com/webstore/detail/lcgbonbkhleeeoibhgepeeeldjmgjhcm">
     <script src="lib/intl/dist/Intl.complete.js"></script>
     <script type="text/javascript">
-      ('Promise' in window && 'reject' in window.Promise) || document.write("<script src='lib/es6-promise/promise.min.js'><\/script>");
+        ('Promise' in window && 'reject' in window.Promise) || document.write("<script src='lib/es6-promise/promise.min.js'><\/script>");
```
arnabk: @4p00rv Could you please check with @ashkanx on this? If we are not supporting browsers which do not have native Promise support, we can handle it in Modernizr check and just recommend users to use the latest version of their browser.

arnabk:
@4p00rv Do we need ach.po/ach.json file for this to work properly?

PR #1036, opened 2017-08-16
Development

4p00rv:
Needs to be merged with [this card](https://trello.com/c/h60hrZ09/2091-fix-binary-style-common).


arnabk:
@4p00rv Already merged it. Anything going to break? When is that other card going to be merged and available

4p00rv:
@arnabk Shouldn't be a problem.

PR #1085, opened 2017-11-14
add compatibility to endpoint chrome extension, add frontend server a…

nazaninreihani:
…s default server

### DIFF
--- a/src/websockets/binary_websockets.es6
+++ b/src/websockets/binary_websockets.es6
@@ -11,7 +11,26 @@ let socket = null;
 let is_website_up = false;
 let queued_requests = {};
 
-const get_app_id = () => window.localStorage.getItem('config.app_id') || 11;
+
+const get_app_default_appid = () => {
+  const app_ids = JSON.parse(app_ids_json);
+  const config = local_storage.get('config');
+  let token = (config && config.app_id) || '';
+
+  if(!token) { /* find the appropriate token */
+    const href = window.location.href;
+    for(const web_address in app_ids) {
+      if(href.lastIndexOf(web_address, 0) == 0) {
+        token = app_ids[web_address];
+        break;
+      }
+    }
+  }
+  localStorage.setItem('config.default_app_id', token);
+  return token;
+};
+
+const get_app_id = () => window.localStorage.getItem('config.app_id') || get_app_default_appid() || 11;
 
 const get_socket_url = () => {
    let server_url = window.localStorage.getItem('config.server_url');
--- a/src/websockets/binary_websockets.es6
+++ b/src/websockets/binary_websockets.es6
@@ -11,7 +11,7 @@ let socket = null;
 let is_website_up = false;
 let queued_requests = {};
 
-const get_app_id = () => window.localStorage.getItem('config.app_id') || (/staging\.binary\.com/i.test(window.location.hostname) ? '1098' : '1');
+const get_app_id = () => window.localStorage.getItem('config.app_id') || 11;
 
 const get_socket_url = () => {
    let server_url = window.localStorage.getItem('config.server_url');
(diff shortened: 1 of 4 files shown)


src/websockets/binary_websockets.es6
```diff
@@ -11,29 +11,24 @@ let socket = null;
...
-
-   if(!token) { /* find the appropriate token */
-      const href = window.location.href;
-      for(const web_address in app_ids) {
-         if(href.lastIndexOf(web_address, 0) == 0) {
-            token = app_ids[web_address];
-            break;
-         }
-      }
+const get_app_id = () => window.localStorage.getItem('config.app_id') || (/staging\.binary\.com/i.test(window.location.hostname) ? '1098' : '1');
```
arnabk: @nazaninreihani I have two questions

1. get_app_id() earlier and get_app_id() that you coded, does not look the same. Looks like they are doing something different.
2. config/config in webtrader was meant to be used by QA to change the endpoint. Now with your new work of allowing QA to change endpoint using chrome ext, I dont think we need config/config module of webtrader.
3. I see you are using 1098 in both next-gen and webtrader. Will it work? App Ids for these apps are different

nazaninreihani: @arnabk You were right, I changed the extension code, so now first I check if the default app_id exists in 'oauth/app_id.json' and if true I set that in storage so the extension can read the default app_id, if the url is not any of the urls mentioned in 'oauth/app_id.json' the default app_id would be the default app_id for webtrader.
just to mention that this extension works only on these urls (containing staging.binary.com, app.binary.com, ticktrade.binary.com, developers.binary.com, www.binary.com and webtrader.binary.com).
which means in localhost the extension is off, (app_id can be changed by changing it in 'oauth/app_id.json'), and the default app_id will be defaults in 'oauth/app_id.json' || webtrader's default. (#1 and #3)

about #2, do you mean I can remove the config file?

src/websockets/binary_websockets.es6
```diff
@@ -11,29 +11,43 @@ let socket = null;
...
+  if(!token) { /* find the appropriate token */
+    const href = window.location.href;
+    for(const web_address in app_ids) {
+      if(href.lastIndexOf(web_address, 0) == 0) {
+        token = app_ids[web_address];
+        break;
       }
+    }
+  }
+  localStorage.setItem('config.default_app_id', token);
```
arnabk: @nazaninreihani In code above you are using `local_storage` and then here you are using `localStorage`?

nazaninreihani: extension needs default_app_id for reset button, 
I have to change the oauth link config too, so I'll fix the localStorage instead of local_storage in that PR

arnabk: @nazaninreihani Couple of points - 
When rest is clicked, just delete the locally stored values and the default hard coded values in the code will take over. I am still not sure why we have to store the default values in local storage.

Using localStorage will cross link the values across applications. Meaning, if you store with config.something using localStorage in webtrader, it might be accessible in other applications. I am not sure if this is desired behavior. We use local_storage in webtrader which ensures that all settings in webtrader/beta or webtrader-live or other applications do not cross link themeselves. You can open applications panel in developer mode and see for yourself.

nazaninreihani: we have so many default_app_id for beta versions and gh-pages of apps that may change in codes by a developer, so It cannot be hardcoded in extension's core codes, and extension needs to access them in local storage to avoid reset to the global default app id. this makes more sense especially in gh-pages, if the default_app_id I've set as a developer in codes is not available to extension it will reset to '1' in binary static for example instead of default app id I've set and makes conflicts between the app and the extension.

arnabk: @nazaninreihani ok, I think before I explain the second paragraph, I probably need to understand the chrome extension throughly. Please see my comment in slack. Let me first get this chrome extension working in my local before going further.

PR #1116, opened 2018-02-23
Hide lookbacks option in Trade select

gokula-krishna-dev:


arnabk:
This PR is not needed. I have disabled it.

PR #1128, opened 2018-04-05
disable reset options

4p00rv:


arnabk:
@4p00rv I recommend submitting this to master because its better to keep this open in beta for testing later when backend is ready for it. No client uses beta, so we are safe.

@ashkanx any comments?

4p00rv:
@arnabk I agree. But let us wait for Ashkan's comment.

ashkanx:
Agree

arnabk:
@4p00rv Closing. Please send a new PR for master

PR #1129, opened 2018-04-05
lookback fix

4p00rv:


arnabk:
@4p00rv Should it be for master too?

4p00rv:
@arnabk No, I will remove the first two commits. This PR is for enabling the lookbacks on webtrader.

PR #1164, opened 2018-07-02
oskar/trackjs_tostring

OskarAhl:
* Add null / undefined check for state.basis.amount
* Fixes error in tradedialog


ashkanx:
@OskarAhl pls solve the conflict

PR #1202, opened 2018-08-06
oskar/real_open_citizen

OskarAhl:
* add mandatory citizen select dropdown, set residence as default


ashkanx:
don't forget to update translation

PR #1237, opened 2018-09-12
oskar/update_contactus_link

OskarAhl:
* Replace redirect function with direct redirect to /contact.html

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -41,7 +41,7 @@
     <div id="topbar" class="topbar primary-bg-color-dark">
         <div class="fr ln29">
             <span class="time"></span>
-            <a target="_blank" href="https://www.binary.com/en/contact.html">
+            <a target="_blank" id="contact-us" href="#">
                 <b>Contact Us</b>
             </a>
             <div class="languages">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -41,7 +41,7 @@
     <div id="topbar" class="topbar primary-bg-color-dark">
         <div class="fr ln29">
             <span class="time"></span>
-            <a target="_blank" href="#" onclick="openContactUs()">
+            <a target="_blank" href="https://www.binary.com/en/contact.html">
```
ashkanx: it's better to redirect user to the respective language instead of EN only.

src/navigation/navigation.html
```diff
@@ -21,7 +21,7 @@
         <li rv-each-ele='languages' class="ele.value" rv-class-invisible="lang.value | eq ele.value" rv-on-click='onclick | bind ele.value'>{ele.name}</li>
       </ul>
     </div>
-    <span class="contact-us"><a href="#" rv-on-click='openContactUs'>Contact Us</a></span>
+    <span class="contact-us"><a href="https://www.binary.com/en/contact.html" target="_blank">Contact Us</a></span>
```
ashkanx: same here

PR #1286, opened 2018-10-31
Amam/issue copy trade

cakasuma:


src/copytrade/copytrade.es6
```diff
@@ -24,6 +24,14 @@ const form_error_messages = {
 
 const getStorageName = () => `copyTrade_${getLoggedInUserId()}`;
 
+const mapCodeToApiCode = (selected_trade_types, all_trade_types) =>
```
OskarAhl: one more thing: can we add a check for if the `selected_trade_types` array is empty

src/copytrade/copytrade.es6
```diff
@@ -282,17 +292,10 @@ const state = {
         .then(tokenUserData => {
           if (!tokenUserData) throw new Error('Invalid token');
           refreshTraderStats(tokenUserData.loginid, scope.searchToken.token, scope)
```
OskarAhl: missing a semicolon here

PR #1288, opened 2018-11-01
Amam/yarn2npm

cakasuma:


package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Webtrader for Binary.com",
   "main": "Gruntfile.js",
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=8.0.0",
+    "yarn": "YARN NO LONGER USED - use npm instead."
```
OskarAhl: I think we can remove this line and update the `README.md` instead

OskarAhl:
We need to update `travis.yml` for this card also

PR #1299, opened 2018-11-05
amam/wrong_pops_up

cakasuma:
fix margin template manager root to avoid wrong pops up in the dialog

### DIFF
--- a/src/trade/tradeDialog.scss
+++ b/src/trade/tradeDialog.scss
@@ -196,6 +196,9 @@
               }
             }
         }
+        .button-disabled {
+          pointer-events: none;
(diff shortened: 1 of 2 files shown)


src/trade/tradeDialog.scss
```diff
@@ -108,7 +108,7 @@
     }
     .trade-template-manager-root {
         position: absolute;
-        margin: -2px 0 0 4px;
+        margin: -2px 0 0 -2px;
```
OskarAhl: I'm still able to open the underlying dropdown through the disabled `save changes`, could you take a look again?

PR #1301, opened 2018-11-05
amam/window_scrolling_issue

cakasuma:
fix resize button asset index


src/assetindex/assetIndex.html
```diff
@@ -1,4 +1,4 @@
-﻿<table class="hover asset-index-dialog">
+﻿<table style="width:100%" class="hover asset-index-dialog">
```
OskarAhl: Can we move this to `assetIndex.scss`? (I know trading times declares it inline but if there is no specific reason for this we should have it in the css instead)

PR #1303, opened 2018-11-05
amam/invalid_time_error

cakasuma:
fix error message on time duration


OskarAhl:
Time picker shouldn't be showing for end time contracts (end time greater than today) mentioned in the trello card as `Also, the time field should be removed when end time is selected(furthermore, it's not editable), as it's not present on binary static page. ` 

You can look at the binary-static implementation for guidance.

PR #1316, opened 2018-11-14
amam/refresh_issue_copy_trade

cakasuma:


src/copytrade/copytrade.es6
```diff
@@ -219,16 +219,16 @@ const state = {
...
+        refreshTraderStats(loginid, token, state)
+        trader.disableRefresh = false;
+        updateLocalStorage(state);
+      }
+      catch (e) {
+        console.error(e)
+        $.growl.error({ message: form_error_messages.REFRESH_FAILED });
+        trader.disableRefresh = false;
+      }
     }
```
OskarAhl: This won't catch async errors since refreshTraderStats is async, I think we can update `refreshTraderStats` to return a promise instead and inside of `refreshTraderStats` resolve it if successful else reject it.

src/copytrade/copytrade.es6
```diff
@@ -282,10 +286,16 @@ const state = {
...
-          updateLocalStorage(scope);
+          refreshTraderStats(tokenUserData.loginid, scope.searchToken.token, scope)
+            .then(() => {
+              scope.searchToken.token = '';
+              scope.searchToken.disable = false;
+              updateLocalStorage(scope);
+            })
+            .catch((e) => {
+              scope.searchToken.disable = false;
+              console.error(e)
```
OskarAhl: Need to add the $growl.error message for users here also

src/copytrade/copytrade.es6
```diff
@@ -282,10 +286,16 @@ const state = {
...
-          updateLocalStorage(scope);
+          refreshTraderStats(tokenUserData.loginid, scope.searchToken.token, scope)
+            .then(() => {
+              scope.searchToken.token = '';
+              scope.searchToken.disable = false;
+              updateLocalStorage(scope);
+            })
+            .catch((e) => {
+              scope.searchToken.disable = false;
+              $.growl.error(e);
```
OskarAhl: This won't work please check the above growl message for when refresh trading statistics request fails.

PR #1318, opened 2018-11-19
amam/fix_window_scrolling

cakasuma:


src/selfexclusion/selfexclusion.es6
```diff
@@ -194,6 +194,7 @@ const init_win = function() {
         });
         win = windows.createBlankWindow($('<div/>'), {
             title: 'Self-Exclusion Facilities'.i18n(),
+            dialogClass: 'selfExclusionDialog',
```
OskarAhl: let's keep the naming of classes with `-` (yes I know some of the dialogClasses are camelCase but it should be with dash

PR #1323, opened 2018-11-22
amam/asset_index_update

cakasuma:


src/assetindex/assetIndex.es6
```diff
@@ -38,163 +31,152 @@ export const init = (li) => {
...
-    });
-    return ret;
+const checkVolatility = (market_name, market_names) => {
+    const volatility_indices = market_names[0][2] ? market_names[0][2].innerText : market_names[0][0].innerText;
+    const is_volatility = market_name.indexOf(volatility_indices) !== -1;
+    return is_volatility;
+}
+
+const getMarketsSubmarkets = (active_symbols) => {
+    const select_market_submarket = active_symbols.reduce((object, item) => {
```
OskarAhl: Can we rename object and item?

src/assetindex/assetIndex.es6
```diff
@@ -38,163 +31,152 @@ export const init = (li) => {
...
 
 const refreshTable = () => {
     const updateTable = (market_name, submarket_name) => {
-        const symbols = markets[market_name][submarket_name];
-        const rows = assets
+        state.dropdown.selected_market = market_name;
+        state.dropdown.is_volatility = checkVolatility(market_name, state.dropdown.display_markets);
+        const symbols = state.dropdown.market_submarkets[market_name][submarket_name];
+        let rows = state.table.asset_data
             .filter((asset) => {
```
OskarAhl: I think we can do this in one line?

src/assetindex/assetIndex.es6
```diff
@@ -38,163 +31,152 @@ export const init = (li) => {
...
+        return rows;
+    }
+    const marketsChanged = (market_submarkets) => {
+        if (!state.dropdown.display_markets) {
+            state.dropdown.display_markets = windows
+                .makeSelectmenu($('<select />').insertBefore(dialog_buttons), {
+                    list: Object.keys(market_submarkets), //Keys are the display_name
+                    inx: 0, //Index to select the item in drop down
+                    changed: (val) => {
+                        const list = Object.keys(market_submarkets[val]); /* get list of sub_markets */
```
OskarAhl: maybe we can call this submarket_list and remove the comment?

src/assetindex/assetIndex.es6
```diff
@@ -38,163 +31,152 @@ export const init = (li) => {
...
+    },
+    table: {
+        asset_data: [],
+        asset_data_extract: [],
+        display_headers: [],
+        display_asset_data: [],
+        search_input: ''
+    },
+    search: function (event, model) {
+        model.table.display_asset_data = [];
```
OskarAhl: model is the same as state here? so perhaps we could call it state.table to make it more readable + const declaration at the top + add space

cakasuma: im not sure which space is it? is it in the array bracket?

PR #1331, opened 2018-11-30
amam/fix_ui_improvements

cakasuma:


src/workspace/workspace.scss
```diff
@@ -65,11 +65,15 @@
...
         line-height: 30px;
         display: flex;
+        cursor: pointer;
         .name {
           padding-left: 24px;
           flex-grow: 1;
           font-size: small;
         }
+        &:hover {
+          background-color: #e8e8e8
```
OskarAhl: Should be a css variable declared at the top

PR #1332, opened 2018-12-04
amam/fix_footer_webtrader

cakasuma:


src/index.js
```diff
@@ -1,74 +1,102 @@
...
-
-        var i18n_name = (window.local_storage.get('i18n') || { value: 'en' }).value;
-        $.getJSON(VERSION + 'i18n/' + i18n_name + '.json', function (data) {
-            setup_i18n_translation(data);
-            populate_footer();
-            if (lang === 'id') {
-                $('#footer').hide();
-            }
-        });
+setSelectedLanguage();
```
OskarAhl: Can we rename this? since the function does more than setting the selectedLanguage. Also I think it makes sense to move `populateLanguageDropdown` inside of this function?

src/index.js
```diff
@@ -80,56 +123,124 @@ function openTradingPage() {
...
-        if (li.value === 'en') li.className = 'invisible';
-        li.innerHTML = '<a href="/" data-lang=' + language.value + '>' + language.name + '</a>';
-        ul_el.appendChild(li);
-    });
-}
+function processFooter() {
+    loadAppId(function(err, app_id_json) {
+        var clients_country;
+        var app_id = err ? default_app_id : getAppId(app_id_json);
+        var i18n_name = (local_storage.get('i18n') || { value: 'en' }).value;
```
OskarAhl: Seems like we are duplicating this logic in 3 places, any way to simplify it?

src/index.js
```diff
@@ -80,56 +123,124 @@ function openTradingPage() {
...
+        var api_url = getUrl() + '?l=' + i18n_name + '&app_id=' + app_id;
+        var ws = new WebSocket(api_url);
```
OskarAhl: Can we add a space after this line?

src/index.js
```diff
@@ -80,56 +123,124 @@ function openTradingPage() {
...
+        ws.onopen = sendWebsiteStatus;
+        ws.onmessage = processApiResponse;
+        function populateFooter(data) {
```
OskarAhl: Can we move this function to the scope where it is being called?

src/index.js
```diff
@@ -80,56 +123,124 @@ function openTradingPage() {
...
+        function sendWebsiteStatus(evt) {
+            sendApiRequest({ website_status: 1 });
+        }
+
+        function sendLandingCompany(clients_country) {
+            sendApiRequest({ landing_company: clients_country });
+        }
+
+        function processApiResponse(response) {
+            var data = JSON.parse(response.data);
```
OskarAhl: can we add error handling here? I think `if (!response.error)` should be enough

src/index.js
```diff
@@ -1,56 +1,83 @@
...
+        }
+    }
+
+    return stored_app_id || default_app_id;
+}
+
+function processPageLanguage() {
+    populateLanguageDropdown();
+
+    if (local_storage.get('oauth') !== null) {
```
OskarAhl: would `if (local_storage.get('oauth')` be enough here?

src/index.js
```diff
@@ -67,8 +94,24 @@ if (local_storage.get('oauth') !== null) {
...
         });
+    }
+}
 
-    });
+function checkRedirectToken(params_str) {
+    if (/acct1=/.test(href) && /token1=/.test(href)) {
+        var accts = (params_str.match(/acct\d=[\w\d]+/g) || []).map(function (val) { return val.split('=')[1] });
+        var tokens = (params_str.match(/token\d=[\w\d-]+/g) || []).map(function (val) { return val.split('=')[1] });
+        var currs = (params_str.match(/cur\d=[\w\d]+/g) || []).map(function (val) { return val.split('=')[1] });
```
OskarAhl: Think we can move these 3 regexes + map to a function to avoid duplication?

PR #1339, opened 2018-12-13
Amam/fix_norway_trading_times

cakasuma:


src/tradingtimes/tradingTimes.es6
```diff
@@ -144,57 +114,74 @@ const initTradingWin = ($html) => {
...
-        }
-
-          updateTable(result, market_names.val(), submarket_names.val());
-          processing_msg.hide();
-         
+               submarket_names.update_list(Object.keys(header[market_names.val()]));
+               submarket_names.off('selectmenuchange', changed);
+               submarket_names.on('selectmenuchange', changed);
+            }
+         }
```
OskarAhl: I think we don't need to make these changes here for now, we can leave it as it is

src/tradingtimes/tradingTimes.es6
```diff
@@ -144,11 +143,13 @@ const initTradingWin = ($html) => {
 
       /* refresh the table with result of {trading_times:yyyy_mm_dd} from WS */
       const refresh = (data) => {
-        const result = processData(menu.extractFilteredMarkets(data));
+        const result = processData(menu.extractFilteredMarkets(data[1]));
+        const header = getMarketsSubmarkets(data[0].active_symbols);
+
```
OskarAhl: Think we can add a `const ... = Object.keys(header)` here and reuse it below at line 161 and 167

src/tradingtimes/tradingTimes.es6
```diff
@@ -172,29 +173,40 @@ const initTradingWin = ($html) => {
...
-      const getCachedData = () => liveapi.cached.send({ trading_times: yyyy_mm_dd })
-      .then(result => refresh(result))
-      .catch((error) => {
-        $.growl.error({ message: error.message });
-        refresh({});
-      });
+      const getCachedData = () => {
+         const active_symbols_request = { active_symbols: 'brief' };
+         const asset_index_request = { trading_times: yyyy_mm_dd };
+         const processing_msg = $('#' + table.attr('id') + '_processing').show();
```
OskarAhl: can we rename this as `$processing_msg` and call `.show()` on the next line to make it more readable

PR #1345, opened 2018-12-14
fix position dialog

cakasuma:


src/windows/windows.scss
```diff
@@ -27,14 +27,20 @@ $COLOR_DARK_BLUE: #2a3052;
             width: fit-content;
         }
 
+        .span-in-dialog-header {
```
OskarAhl: can we group these three classes inside `.trading-times.span-in-dialog-header, statement..., .profitTable...`?

cakasuma: should be possible

PR #1353, opened 2018-12-21
amam/fix_pa_withdraw_crypto

cakasuma:


src/cashier/withdraw.es6
```diff
@@ -158,10 +161,20 @@ class Withdraw {
...
             menu.choice = choice;
             route.update(choice !== 'transfer' ? 'verify' : 'transfer');
+
             if (choice === 'transfer')
                 return;
+
             const email = local_storage.get('authorize').email;
             const type = choice === 'agent' ? 'paymentagent_withdraw' : 'payment_withdraw';
+
+            isCrypto() && (agent.currency !== 'DAI' || agent.currency !== 'UST') ? (
```
morteza-binary: @cakasuma It would be more clear to use `if/else` instead of ternary operation since there isn't any value assignment and the condition in too long.

PR #1358, opened 2018-12-26
amam/trading_asset_reverse

cakasuma:
- handle the incorrect arrangement of asset index and trading times
- reduce calling active symbols and trading times caching

### DIFF
--- a/src/assetindex/assetIndex.es6
+++ b/src/assetindex/assetIndex.es6
@@ -81,7 +81,6 @@ const initTable = () => {
 
     function populateTable(result) {
         const active_symbols_data = local_storage.get('active_symbols');
-        console.log(result[0].asset_index);
         const asset_index_data = [...result[0].asset_index];
 
         if($.isEmptyObject(active_symbols_data) && $.isEmptyObject(asset_index_data)) return;
--- a/src/assetindex/assetIndex.es6
+++ b/src/assetindex/assetIndex.es6
@@ -84,6 +84,7 @@ const initTable = () => {
         const asset_index_data = [...result[1].asset_index];
 
         if($.isEmptyObject(active_symbols_data) && $.isEmptyObject(asset_index_data)) return;
+
         state.dropdown.market_submarkets = getMarketsSubmarkets(active_symbols_data);
         state.dropdown.sorted_markets = getDefaultMarkets(active_symbols_data);
         state.table.asset_data = asset_index_data;
--- a/src/assetindex/assetIndex.es6
+++ b/src/assetindex/assetIndex.es6
@@ -4,7 +4,7 @@ import liveapi from '../websockets/binary_websockets';
 import rv from 'common/rivetsExtra';
 import 'jquery-growl';
 import 'css!./assetIndex.css';
-import getMarketsSubmarkets from '../common/marketUtils';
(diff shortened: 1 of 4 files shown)


src/assetindex/assetIndex.es6
```diff
@@ -155,10 +155,13 @@ const initTable = () => {
         }
 
         function marketsDropdown(market_submarkets) {
+            const markets_array = Object.keys(market_submarkets);
+            const markets_reversed = markets_array.length > 1 ? Object.assign([], markets_array, {[2]: markets_array[3], [3]: markets_array[2]}) : markets_array; // handle norway clients (handle translation)
```
OskarAhl: I think it's possible to add a function for `markets_reversed` and add it to marketUtils?

src/tradingtimes/tradingTimes.es6
```diff
@@ -146,6 +146,7 @@ const initTradingWin = ($html) => {
         const result = processData(menu.extractFilteredMarkets(data[1]));
         const header = getMarketsSubmarkets(data[0].active_symbols);
         const market_name_list = Object.keys(header);
+        const market_name_reversed = market_name_list.length > 1 ? Object.assign([], market_name_list, {[2]: market_name_list[3], [3]: market_name_list[2]}) : market_name_list; // handle norway clients (handle translation)
```
OskarAhl: and use the function here too (to keep things dry 😄 )

cakasuma: indeed

negarn:
conflict

src/common/marketUtils.es6
```diff
@@ -12,4 +12,6 @@ const getMarketsSubmarkets = (active_symbols) => {
     return select_market_submarket
  }
 
-export default getMarketsSubmarkets;
\ No newline at end of file
+const getSortedMarkets = (markets) => markets.length > 1 ? Object.assign([], markets, {[2]: markets[3], [3]: markets[2]}) : markets;
```
OskarAhl: How is this function sorting the markets array? perhaps we could include the purpose of the function in either the name and/or parameters?

cakasuma: because the returned array from the api is always vol indices first before commodities, that functions swapping the position between them both. as for if its norway clients where they only have volatility indices then it will stay as it is.

src/assetindex/assetIndex.es6
```diff
@@ -79,12 +80,14 @@ const initTable = () => {
         });
 
     function populateTable(result) {
-        const active_symbols_data = Object.assign(result[0].active_symbols);
-        const asset_index_data = [...result[1].asset_index];
+        const active_symbols_data = local_storage.get('active_symbols');
+        console.log(result[0].asset_index);
```
OskarAhl: We can remove this console.log

src/common/marketUtils.es6
```diff
@@ -1,15 +1,43 @@
+const market_order = {
+    forex: 1,
+    indices: 2,
+    commodities: 3,
+    volidx: 4,
+};
+
+const getMarketPosition = (markets) => markets.sort((a, b) => market_order[a] - market_order[b]);
```
OskarAhl: Since `market_order` is only used in `getMarketPosition `, maybe we can hide it with a closure within `getMarketPosition `?

PR #1362, opened 2019-01-04
amam/fix_round_up_barrier_2

cakasuma:
## fix tradeConf dialog
- add fractional digits of display decimals to the buy barrier and remove add comma in html

### DIFF
--- a/src/trade/tradeConf.es6
+++ b/src/trade/tradeConf.es6
@@ -155,7 +155,7 @@ const registerTicks = (state, extra) => {
 
       if (should_add_new_tick) {
          const contract_is_finished = proposal_open_contract.status !== 'open' && (tick_count < -1 || is_path_dependent_last_tick);
-         state.buy.barrier = proposal_open_contract.barrier ? addComma(+proposal_open_contract.barrier, display_decimals) : null;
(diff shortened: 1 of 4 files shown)


src/trade/tradeConf.es6
```diff
@@ -155,7 +155,7 @@ const registerTicks = (state, extra) => {
 
       if (should_add_new_tick) {
          const contract_is_finished = proposal_open_contract.status !== 'open' && (tick_count < -1 || is_path_dependent_last_tick);
-         state.buy.barrier = proposal_open_contract.barrier ? (+proposal_open_contract.barrier) : null;
+         state.buy.barrier = proposal_open_contract.barrier ? addComma(+proposal_open_contract.barrier, display_decimals) : null;
```
OskarAhl: This will cause the barrier to not be drawn on the chart. One option would be to add display_decimals to the state and pass it to the `add-comma` rivets formatter

src/trade/tradeConf.es6
```diff
@@ -312,6 +312,7 @@ export const init = (data, extra, showCallback, hideCallback) => {
       ticks: {
          array: [],
          contract_is_finished: false,
+         display_decimals: display_decimals,
```
OskarAhl: nit: I think `display_decimals,` is enough since the property is the same name

PR #1363, opened 2019-01-04
oskar/add_crypto_basis_decimals

OskarAhl:
* set separate basis_decimals for payout and stake for crypto based on `fractional_digits ` from website_status call (except for lookbacks - 3 decimals)

### DIFF
--- a/src/trade/tradeDialog.es6
+++ b/src/trade/tradeDialog.es6
@@ -291,14 +291,13 @@ function init_state(available,root, dialog, symbol, contracts_for_spot){
     currency: {
       array: ['USD'],
       value: 'USD',
-      decimals: 0,
+      decimals: currencyFractionalDigits(),
     },
     basis: {
       array: ['Payout', 'Stake'],
       value: 'payout',
       amount: currencyFractionalDigits() === 8 ? 0.10000000 : 10,
       limit: null,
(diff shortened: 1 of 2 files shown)


src/trade/tradeDialog.es6
```diff
@@ -379,8 +379,8 @@ function init_state(available,root, dialog, symbol, contracts_for_spot){
       name: '',
       visible: false,
     },
+    basis_decimals: isCryptoCurrency(local_storage.get('currency')) ? 8 : 2,
   };
-
```
cakasuma: i think its better to keep the space here ?

src/trade/tradeDialog.es6
```diff
@@ -379,8 +379,8 @@ function init_state(available,root, dialog, symbol, contracts_for_spot){
       name: '',
       visible: false,
     },
+    basis_decimals: isCryptoCurrency(local_storage.get('currency')) ? 8 : 2,
```
cakasuma: is the crypto decimals always 8 ?

cakasuma: instead of adding new object, we can put it under `basis` object on line 296 ?, so basis.decimals

cakasuma:
i think currency.decimals in the state have the same value which is currencyFractionalDigits()

PR #1377, opened 2019-01-29
amam/fix_trade_menu_order

cakasuma:


src/navigation/menu.es6
```diff
@@ -59,8 +55,8 @@ export const sortMenu = (markets) => {
    //Sort market
    if($.isArray(markets)) {
 
-      const rank = { "forex": 1, "indices": 2, "otc stocks": 3, "commodities": 4, "volatility indices": 5 };
-      markets = _.sortBy( markets, (o) => rank[o.display_name.toLowerCase()]);
+      const rank = { "forex": 1, "indices": 2, "otc stocks": 3, "commodities": 4, "volidx": 5 };
+      markets = _.sortBy( markets, (o) => rank[o.name.toLowerCase()]);
```
OskarAhl: Can't we use our marketUtils in common for this?

cakasuma: yes im planning to do so as well 👍 , but if need quick fix, then this is the fastest solution

PR #1387, opened 2019-02-04
amam/upgrade_dependencies_deployment

cakasuma:
- remove grunt if (contains high vulnerability dependencies)
- register task `deploy` for conditional deployment


Gruntfile.js
```diff
@@ -403,4 +379,13 @@ module.exports = function (grunt) {
     grunt.registerTask("deploy-branch", ["default","gitinfo", "clean:current_branch", "copy:copy_current_branch", "gh-pages:deploy-branch"]);
     /* clean all the files in gh-pages branch */
     grunt.registerTask("gh-pages-clean", ["gh-pages:clean"]);
+
+    // conditional switch for deployment
+    grunt.registerTask("deploy", function (deployment_mode) {
```
OskarAhl: I think we don't need the `deployment_mode` parameter, checking the `TRAVIS_BRANCH` should be enough

package.json
```diff
@@ -43,7 +42,7 @@
     "lodash": "^4.17.5"
   },
   "scripts": {
-    "deploy-gh-pages": "grunt if:live if:beta",
+    "deploy-gh-pages": "grunt deploy:live deploy:beta",
```
OskarAhl: And then we can just have `grunt deploy` here

PR #1394, opened 2019-02-18
Amam/fix dialog overlapping

cakasuma:
fix dialog overlapping and end time not drawn intermittenly

### DIFF
--- a/src/viewtransaction/viewTransaction.es6
+++ b/src/viewtransaction/viewTransaction.es6
@@ -350,7 +350,7 @@ function drawXLines(contract, state, chart) {
     }
 
     if (is_sold_before_expiry) drawXLine({ line_time: sell_time, label: 'end_time', dashStyle: 'Dash' });
-    if (!is_path_dependent) drawXLine({ line_time: date_expiry, label: 'end_time', dashStyle: 'Dash' });
+    if (!is_path_dependent) drawXLine({ line_time: date_expiry && exit_tick_time, label: 'end_time', dashStyle: 'Dash' });
   }
 
   function drawPurchaseTime({ purchase_time }) {
--- a/src/viewtransaction/viewTransaction.es6
+++ b/src/viewtransaction/viewTransaction.es6
@@ -350,7 +350,7 @@ function drawXLines(contract, state, chart) {
     }
 
     if (is_sold_before_expiry) drawXLine({ line_time: sell_time, label: 'end_time', dashStyle: 'Dash' });
-    if (!is_path_dependent && date_expiry) drawXLine({ line_time: date_expiry, label: 'end_time', dashStyle: 'Dash' });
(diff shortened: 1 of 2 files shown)


src/viewtransaction/viewTransaction.es6
```diff
@@ -350,7 +350,7 @@ function drawXLines(contract, state, chart) {
     }
 
     if (is_sold_before_expiry) drawXLine({ line_time: sell_time, label: 'end_time', dashStyle: 'Dash' });
-    if (!is_path_dependent) drawXLine({ line_time: date_expiry, label: 'end_time', dashStyle: 'Dash' });
+    if (!is_path_dependent) drawXLine({ line_time: date_expiry && exit_tick_time, label: 'end_time', dashStyle: 'Dash' });
```
OskarAhl: shouldn't it be `if (!is_path_dependent && date_expiry)`?

cakasuma: indeed. when i debug it the problem was, every tick should have date_expiry carry along. but then some ticks are not consistent when it stops and date expiry is not exist. thats when the problem happens. it depends on internet connection, sometime i got all date expiry carry along, sometimes some tick dont have date expiry. but by default, they should have date expiry. so i made a fallback if date expiry is not exist then use exit tick time instead. what do you think

src/viewtransaction/viewTransaction.es6
```diff
@@ -350,7 +350,7 @@ function drawXLines(contract, state, chart) {
     }
 
     if (is_sold_before_expiry) drawXLine({ line_time: sell_time, label: 'end_time', dashStyle: 'Dash' });
-    if (!is_path_dependent) drawXLine({ line_time: date_expiry, label: 'end_time', dashStyle: 'Dash' });
+    if (!is_path_dependent) drawXLine({ line_time: (date_expiry || exit_tick_time), label: 'end_time', dashStyle: 'Dash' });
```
OskarAhl: I think we always want to use the `exit_tick_time` as basis for end_time, if response doesn't have `exit_tick_time` then better to wait with drawing the end time until `exit_tick_time` is received

cakasuma: `drawXLines` is always executed when the tick comes. for days contract, we need to use date_expiry as end time

OskarAhl: It's drawn based on the `proposal_open_contract` stream / response, I think days also have exit_tick_time?

cakasuma: yes, they have exit tick time, but, i dont think we can draw exit tick time for end time except path dependent contract. bcause in binary static, its totally the date expiry. i think i have to find a way. to make it draw the date_expiry regardless it exist or not.

cakasuma: one solution might be triggering to draw chart after the contract finish

PR #1395, opened 2019-02-18
Amam/ifx endtime expiry remaining

cakasuma:


src/common/rivetsExtra.es6
```diff
@@ -432,6 +432,10 @@ rv.binders.slider = {
    }
 }
 
+rv.binders.mindatepicker = {
```
OskarAhl: `datepicker_mindate` would be easier to read?

src/trade/tradeDialog.es6
```diff
@@ -580,19 +581,29 @@ function init_state(available,root, dialog, symbol, contracts_for_spot) {
...
+    const is_daily_contracts = state.duration_unit.array[0] && !hasIntradayUnit(state.duration_unit.array);
+
     if (!is_today) {
       state.date_expiry.today_times.disabled = true;
         trading_times_for(value_date, state.proposal.symbol)
           .then(function(times){
+            state.date_expiry.min_date = is_daily_contracts ? 1 : 0;
             const value_hour = times.close !== '--' ? times.close : '23:59:59';
             state.date_expiry.value_hour = moment.utc(value_hour, 'HH:mm:ss').format('HH:mm');
+            if (state.duration.value === 'Duration') {
```
OskarAhl: I don't think this will work with a translated 'Duration'?

cakasuma: the `state.duration.value` taking the `HH:mm` format for endtime and when it goes duration it will changed to duration. i thought its the code for duration. let me check it again later

cakasuma: i just check in different languages, the `state.duration.value` remain same which is `End time` and `Duration`

src/trade/tradeDialog.es6
```diff
@@ -913,6 +924,10 @@ function init_state(available,root, dialog, symbol, contracts_for_spot) {
     return barrier && (barrier.startsWith('+') || barrier.startsWith('-')) ? true : false;
   }
 
+  function hasIntradayUnit(duration_unit_array) {
```
OskarAhl: Could you also move this above `function init_state(...)`? let's keep all the component specific helper functions there

src/trade/tradeDialog.es6
```diff
@@ -913,6 +930,8 @@ function init_state(available,root, dialog, symbol, contracts_for_spot) {
     return barrier && (barrier.startsWith('+') || barrier.startsWith('-')) ? true : false;
   }
 
+
+
```
OskarAhl: extra blank lines? 😄

PR #1400, opened 2019-02-22
amam/dropdown_mismatch

cakasuma:
add sorted markets function in market utils to sort objects

### DIFF
--- a/src/common/marketUtils.es6
+++ b/src/common/marketUtils.es6
@@ -11,7 +11,7 @@ const getMarketPosition = (() => {
     };
 })();
 
-const getMarketChild = (markets) => {
+const getMarketChild = ([...markets]) => {
     return markets.map((market) => {
         if (market.submarkets) {
             const submarkets_sort = market.submarkets.sort((a, b) => {
@@ -61,7 +61,7 @@ const getMarkets = (active_symbols) => {
     }, {});
 };
 
-function sortSubmarkets(submarkets) {
+function sortSubmarkets([...submarkets]) {
     return submarkets.sort((a, b) => {
         if (a < b) return -1;
         if (a > b) return 1;
--- a/src/common/marketUtils.es6
+++ b/src/common/marketUtils.es6
@@ -1,4 +1,4 @@
-const getMarketPosition = (() => {
+const sortMarkets = (() => {
     const market_order = {
         forex: 1,
         indices: 2,
@@ -11,7 +11,7 @@ const getMarketPosition = (() => {
     };
 })();
 
-const sortMarketSubmarkets = ([...markets]) => {
+const sortSubmarkets = ([...markets]) => {
     return markets.map((market) => {
         if (market.submarkets) {
             const submarkets_sort = market.submarkets.sort((a, b) => {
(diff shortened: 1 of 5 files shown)


src/common/marketUtils.es6
```diff
@@ -11,6 +11,26 @@ const getMarketPosition = (() => {
...
 })();
 
+const getMarketChild = (markets) => {
+    return markets.map((market) => {
+        if (market.submarkets) {
+            const submarkets_sort = market.submarkets.sort((a, b) => {
+                if (a.display_name < b.display_name) return -1;
+                if (a.display_name > b.display_name) return 1;
+            });
+            market.submarkets = submarkets_sort;
```
OskarAhl: I think this will mutate the market passed in as a argument, to decouple and avoid side-effects better to:

`const getMarketChild = ([...markets]) => {`

src/common/marketUtils.es6
```diff
@@ -41,8 +61,16 @@ const getMarkets = (active_symbols) => {
     }, {});
 };
 
+function sortSubmarkets(submarkets) {
```
OskarAhl: For good practice let's do the same here: `sortSubmarkets([...submarkets])`

src/common/marketUtils.es6
```diff
@@ -11,6 +11,26 @@ const getMarketPosition = (() => {
     };
 })();
 
+const getMarketChild = ([...markets]) => {
```
OskarAhl: Maybe sortMarketSubmarkets is more descriptive?

src/common/marketUtils.es6
```diff
@@ -41,8 +61,16 @@ const getMarkets = (active_symbols) => {
     }, {});
 };
 
+function sortSubmarkets([...submarkets]) {
+    return submarkets.sort((a, b) => {
+        if (a < b) return -1;
```
OskarAhl: Why is the above sort sorting by display_name and not this one?

cakasuma: `getMarketChild` is sorting the submarket as a whole object, used in header dropdown. however, `sortSubmarket` is sorting submarket as an array, used in asset index and trading times dropdown.

OskarAhl: Ah I see 👍

src/common/marketUtils.es6
```diff
@@ -11,6 +11,26 @@ const getMarketPosition = (() => {
...
+                if (a.display_name < b.display_name) return -1;
+                if (a.display_name > b.display_name) return 1;
+            });
+            market.submarkets = submarkets_sort;
+        }
+        return market
+    })
+}
+
+const getMarketChildSorted = (markets) => {
```
OskarAhl: This function sorts markets and submarkets?

cakasuma: yes, that is the compilation of sorting the object, which sorts market and submarkets. which we will export

cakasuma: but the name should be change to `getMarketSubmarketsSorted` i think

cakasuma: let me rename all the function to be more descriptive

PR #1405, opened 2019-02-26
update include list contract

cakasuma:


src/trade/tradeDialog.es6
```diff
@@ -248,6 +248,8 @@ function hasIntradayUnit(duration_unit_array) {
 }
 
 function init_state(available,root, dialog, symbol, contracts_for_spot) {
+  const include_list = ['up/down', 'touch/no touch', 'in/out', 'digits', 'asians', 'rise/fall equal', 'lookbacks'];
```
OskarAhl: What about `SUPPORTED_CONTRACT_TYPES`? we could add this to a config file in common - think we need it anywhere else? And better to wrap it in `Object.freeze()` so it can't be changed.

PR #1418, opened 2019-03-04
amam/fix_xss_bug

cakasuma:


Shashank-In:
For safer side can we filter out brackets as well "(" and ")"
There are certain bypasses which can work without angle brackets and quotes.

PR #1421, opened 2019-03-05
amam/upload_xml_bug

cakasuma:
- sanitizing all possible workspace name (especially on load)
- add on keypress sanitization

### DIFF
--- a/src/workspace/workspace.es6
+++ b/src/workspace/workspace.es6
@@ -51,16 +51,16 @@ const state = {
             tracker.reopen(clone(w));
          }, 500);
       },
-      perv_name: '',
-      save_name: w => state.workspace.perv_name = sanitize(w.name),
+      prev_name: '',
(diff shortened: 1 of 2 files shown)


src/workspace/workspace.es6
```diff
@@ -55,13 +55,13 @@ const state = {
       save_name: w => state.workspace.perv_name = sanitize(w.name),
       blur: el => el.blur(),
       rename: w => {
-        const perv_name = state.workspace.perv_name;
+        const perv_name = sanitize(state.workspace.perv_name);
```
OskarAhl: We might as well update this to `prev_name` also since we are changing it + the other `pervs` also 😉

cakasuma: there are quite number of `pervs` in other files too. for now i only change `pervs` in workspace.

PR #1424, opened 2019-03-27
Amam/navbar hover fix

cakasuma:
- add description to the html element
- click menu instead of hover menu to show menu dialog


src/navigation/navigation.es6
```diff
@@ -8,6 +8,13 @@ import workspace from '../workspace/workspace.js';
 import '../common/util';
 import 'css!navigation/navigation.css';
 
+const menu_selectors = [
+   '.trade',
+   '.instruments',
+   '.resources',
+   '.workspace',
+]
```
OskarAhl: missing `;`

PR #1429, opened 2019-04-02
Khalid/added 18+ underage sign to footer

Khalidxx:
Also, a small change to the URL for requesting style files


package.json
```diff
@@ -15,7 +15,7 @@
     "babel-preset-stage-0": "^6.5.0",
     "concurrently": "^3.3.0",
     "connect": "^3.6.1",
-    "grunt": "^1.0.1",
```
cakasuma: can do this in seperate PR

src/index.scss
```diff
@@ -140,6 +143,10 @@ body {
         > * {
             margin-right: 20px;
         }
+        @media screen and (max-width: 800px)  {
```
cakasuma: i think we can combine these two media queries since it is the same pixel

package-lock.json
```diff
@@ -1892,10 +1892,10 @@
       "integrity": "sha1-DQcLTQQ6W+ozovGkDi7bPZpMz3c=",
```
OskarAhl: Can we move the `package-lock.json` changes to a separate card?

Khalidxx: Alright, resolved

OskarAhl: Thanks 👍

PR #1433, opened 2019-04-05
khalid/removed google+ from footer

Khalidxx:


src/index.scss
```diff
@@ -141,11 +141,13 @@ body {
     &__regulatory-icons {
         display: flex;
         align-items: center;
+        max-width: 400px;
```
cakasuma: this can be `width: fit-content`

Khalidxx: But it doesn't work very well with tablets (less than 800px)

PR #1442, opened 2019-04-29
Amam/cached_liveapi

cakasuma:
liveapi send change to liveapi cached send too many call same request
call cached:
`active_symbols`
`ticks` related


OskarAhl:
Which calls are sending too many of the same request? I don't really understand the point of this, can you explain more please.

cakasuma:
> Which calls are sending too many of the same request? I don't really understand the point of this, can you explain more please.

i found a lot of trackJS error regarding the `"code":"WrongResponse","message":"Sorry, an error occurred while processing your request.` . i think its always better to cached it rather than always sending the new one in this case that is when multiple modules request
the same data or a module request a data multiple times. @OskarAhl what do you think?

PR #1468, opened 2019-05-27
khalid/fixed dragging issues

Khalidxx:
- Dragging containment disabled after a window is maximiszd and then restored
- Dragging limit is set to disable dragging horizontally outside the window

### DIFF
--- a/src/windows/tracker.es6
+++ b/src/windows/tracker.es6
@@ -225,9 +225,7 @@ export const track = (options, blankWindow) => {
       save_states();
    });
    blankWindow.on('dialogextendrestore',() => {
-      dialog.draggable({
-         containment: false,
-      })
+      dialog.draggable({ containment: false })
       state.position.offset = dialog.offset();
       state.position.mode = 'normal';
       save_states();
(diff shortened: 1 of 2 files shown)


src/windows/tracker.es6
```diff
@@ -225,6 +225,9 @@ export const track = (options, blankWindow) => {
       save_states();
    });
    blankWindow.on('dialogextendrestore',() => {
+      dialog.draggable({
```
cakasuma: can make this one line

src/windows/windows.es6
```diff
@@ -331,13 +331,18 @@ export const createBlankWindow = function($html,options) {
    dialog.on('dragstop', () => {
       const top = dialog.offset().top;
       const left = dialog.offset().left;
+      const dialogWidth = blankWindow.dialog( "option", "width" );
+      const windowWidth = $(window).width();
 
-      if (top < 0) {
-         dialog.animate({ top: '0px' }, 300, dialog.trigger.bind(dialog, 'animated'));
+      if (top < 106) {
```
cakasuma: can we get nav height instead of hardcoded height value

PR #1469, opened 2019-05-31
Khalid/transactions bug

Khalidxx:
- changed conditions to enable `value` of 0 to be a valid digit
- fixed missing `Final price`
- fixed `Barrier` of 0 issue

### DIFF
--- a/src/common/rivetsExtra.es6
+++ b/src/common/rivetsExtra.es6
@@ -61,7 +61,7 @@ rv.formatters['gt'] = (vlaue, other) => vlaue > other;
 /* rivets formatter for < operator  */
 rv.formatters['lt'] = (vlaue, other) => vlaue < other;
 /* localise price format*/
-rv.formatters['format-price'] = (value, currency) => (value) ?  formatPrice(value, currency) : undefined;
+rv.formatters['format-price'] = (value, currency) => (value != undefined) ?  formatPrice(value, currency) : undefined;
 /* comma added format*/
 rv.formatters['add-comma'] = (value, decimal_points) => (value) ?  addComma(value, decimal_points) : undefined;
 /* rivets formater to capitalize string */
--- a/src/common/rivetsExtra.es6
+++ b/src/common/rivetsExtra.es6
@@ -63,7 +63,7 @@ rv.formatters['gt'] = (vlaue, other) => vlaue > other;
 /* rivets formatter for < operator  */
 rv.formatters['lt'] = (vlaue, other) => vlaue < other;
 /* rivets formatter to check if a variable has value */
-rv.formatters['has-value'] = (value) => hasValue(value);
+rv.formatters['has-value'] = hasValue;
 /* localise price format*/
 rv.formatters['format-price'] = (value, currency) => hasValue(value) ? formatPrice(value, currency) : undefined;
 /* comma added format*/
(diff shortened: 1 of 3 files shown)


src/common/rivetsExtra.es6
```diff
@@ -60,10 +60,12 @@ rv.formatters['and-not'] = (vlaue, other) => vlaue && !other;
 rv.formatters['gt'] = (vlaue, other) => vlaue > other;
 /* rivets formatter for < operator  */
 rv.formatters['lt'] = (vlaue, other) => vlaue < other;
+/* rivets formatter to check if a variable has value */
+rv.formatters['has-value'] = (value) => value != null;
```
OskarAhl: `undefined`, `NaN` will still pass this check. 

Since you need a check for "truthy" numbers I think you could change this to `has-number` and find a way for it to check if the number is "truthy" and not 0

https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/common/rivetsExtra.es6
```diff
@@ -151,7 +153,7 @@ rv.formatters['bold-last-character'] = (str) => {
 }
 /* formatter to calcualte the percent of a value of another value */
 rv.formatters['percent-of'] = (changed, original) => {
-   if(!changed || !original) return undefined;
+   if(!(changed != null) || !original) return undefined;
```
OskarAhl: missing space `if (`

src/common/rivetsExtra.es6
```diff
@@ -60,10 +60,12 @@ rv.formatters['and-not'] = (vlaue, other) => vlaue && !other;
 rv.formatters['gt'] = (vlaue, other) => vlaue > other;
 /* rivets formatter for < operator  */
 rv.formatters['lt'] = (vlaue, other) => vlaue < other;
+/* rivets formatter to check if a variable has value */
+rv.formatters['has-value'] = (value) => value != null;
 /* localise price format*/
-rv.formatters['format-price'] = (value, currency) => (value) ?  formatPrice(value, currency) : undefined;
+rv.formatters['format-price'] = (value, currency) => (value != null) ?  formatPrice(value, currency) : undefined;
```
OskarAhl: same here

src/common/rivetsExtra.es6
Khalidxx: `undefined` won't pass this (since it's one `=`), but I don't know what to do with `NaN` values without being verbose. Maybe I can do something like this:

```javascript
rv.formatters['has-value'] = (value) => value === 0 ? true : Boolean(value)
```

src/common/rivetsExtra.es6
Khalidxx: I think we can use `isNaN` here, because the function is specific to numbers. What do you think?

src/common/rivetsExtra.es6
```diff
@@ -60,18 +60,20 @@ rv.formatters['and-not'] = (vlaue, other) => vlaue && !other;
 rv.formatters['gt'] = (vlaue, other) => vlaue > other;
 /* rivets formatter for < operator  */
 rv.formatters['lt'] = (vlaue, other) => vlaue < other;
+/* rivets formatter to check if a variable has value */
+rv.formatters['has-value'] = (value) => value === 0 ? true : Boolean(value);
 /* localise price format*/
-rv.formatters['format-price'] = (value, currency) => (value) ?  formatPrice(value, currency) : undefined;
+rv.formatters['format-price'] = (value, currency) => (!isNaN(value)) ?  formatPrice(value, currency) : undefined;
```
OskarAhl: Can we use  Number.isNaN() and I think we still need to check `(value) ?  formatPrice(value, currency) : ...`?

Khalidxx: Okay, changed it to a new function

PR #1471, opened 2019-06-11
khalid/statement issues

Khalidxx:
Issues fixed:
- Showing barrier of `-` instead of `0` for `Match/Differ` and `Over/Under` contracts
- Showing multiple `rateLimit` error messages when scrolling to the end of Statements window

### DIFF
--- a/src/common/rivetsExtra.es6
+++ b/src/common/rivetsExtra.es6
@@ -67,7 +67,7 @@ rv.formatters['has-value'] = hasValue;
 /* localise price format*/
 rv.formatters['format-price'] = (value, currency) => hasValue(value) ? formatPrice(value, currency) : undefined;
 /* comma added format*/
-rv.formatters['add-comma'] = (value, decimal_points) => (parseFloat(value)) ?  addComma(value, decimal_points) : undefined;
+rv.formatters['add-comma'] = (value, decimal_points) => value ?  addComma(value, decimal_points) : undefined;
 /* rivets formater to capitalize string */
 rv.formatters.capitalize = {
    read: (value) => _.capitalize(value),
--- a/src/common/rivetsExtra.es6
+++ b/src/common/rivetsExtra.es6
@@ -67,7 +67,7 @@ rv.formatters['has-value'] = hasValue;
 /* localise price format*/
 rv.formatters['format-price'] = (value, currency) => hasValue(value) ? formatPrice(value, currency) : undefined;
 /* comma added format*/
-rv.formatters['add-comma'] = (value, decimal_points) => hasValue(value) ?  addComma(value, decimal_points) : undefined;
+rv.formatters['add-comma'] = (value, decimal_points) => value ?  addComma(value, decimal_points) : undefined;
 /* rivets formater to capitalize string */
 rv.formatters.capitalize = {
    read: (value) => _.capitalize(value),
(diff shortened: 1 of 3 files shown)


src/statement/statement.es6
```diff
@@ -89,6 +90,7 @@ const refreshTable  = (yyy_mm_dd) => {
             .then(refresh)
             .catch((err) => {
                   refresh({});
+                  api_limit_reached = true;
```
OskarAhl: Should this be reset after one minute? (since the rate limit will be cleared by then)

src/common/rivetsExtra.es6
```diff
@@ -67,7 +67,7 @@ rv.formatters['has-value'] = hasValue;
 /* localise price format*/
 rv.formatters['format-price'] = (value, currency) => hasValue(value) ? formatPrice(value, currency) : undefined;
 /* comma added format*/
-rv.formatters['add-comma'] = (value, decimal_points) => (parseFloat(value)) ?  addComma(value, decimal_points) : undefined;
+rv.formatters['add-comma'] = (value, decimal_points) => value ?  addComma(value, decimal_points) : undefined;
```
OskarAhl: Since value can be a number, we could add the `hasValue` check here also?

Khalidxx: Yeah, that should be better

src/viewtransaction/viewTransaction.es6
```diff
@@ -233,11 +239,11 @@ const updateIndicative = (data, state) => {
     state.proposal_open_contract.current_spot = contract.current_spot;
     state.proposal_open_contract.current_spot_time = contract.current_spot_time;
     state.proposal_open_contract.bid_price = contract.bid_price;
-    state.proposal_open_contract.entry_tick = contract.entry_tick;
+    state.proposal_open_contract.entry_tick = countDecimals(contract.entry_tick) < 3 ? contract.entry_tick.toFixed(3) : contract.entry_tick;
```
OskarAhl: Why is this necessary?

src/viewtransaction/viewTransaction.es6
```diff
@@ -45,6 +45,12 @@ const showMarketDataDisruptionWindow = () => {
    window.dd = market_data_disruption_win;
 };
 
+const countDecimals = (value) => { 
+  if ((value % 1) != 0)
```
OskarAhl: Should be `!==`

src/viewtransaction/viewTransaction.es6
Khalidxx: It's because the entry/exit spot doesn't show enough decimal places. For example `123.45` is supposed to be shown `123.450`. Sometimes they choose a `match/differ` contract and last digit prediction as 5, which means they lost in this case, but they see last digit as 5. They will get really angry and complain 😂

src/viewtransaction/viewTransaction.es6
Khalidxx: Yes, for sure

src/viewtransaction/viewTransaction.es6
OskarAhl: 👍 ah yes, I think we can use `display_decimals` instead of `3`?

Khalidxx: Oh okay, I'll change it

PR #1488, opened 2019-08-08
update binary-style to v0.2.18

brandon-binary:


package-lock.json
```diff
@@ -7049,17 +7057,6 @@
       "dev": true,
       "optional": true
     },
-    "shelljs": {
```
mustofa-binary: why the latest shelljs get deleted?

PR #1518, opened 2020-02-04
shudarshon.chaks/enabled artifact and artifact checksum upload steps in circleci

shudarshon-binary:


.circleci/config.yml
```diff
@@ -87,6 +122,9 @@ jobs:
       - git_checkout_from_cache
       - npm_install
       - build
+      - compress
+      - upload_artifact
```
4p00rv:
```suggestion
       # Uploading the built code to s3 to create a backup of key services separate from Kubernetes deployment.
      - upload_artifact
```
Adding a comment to let the future developers know of the reason for this.

deploy/cron.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
4p00rv: @shudarshon-binary I don't think we need this file in the repo.

.circleci/config.yml
```diff
@@ -52,18 +53,52 @@ commands:
...
+          name: "Compile project"
+          command: npm run build
+
+  compress:
+    description: "Compress"
+    steps:
+      - run:
+          name: "Compress"
+          command: |
+            pushd dist/compressed/
```
aminmarashi-binary: nitpicking: `pushd` without a `popd`. Better to use `cd`.

aminmarashi-binary: I see, used `OLDPWD` instead. :ok_hand:

.circleci/config.yml
```diff
@@ -52,18 +53,52 @@ commands:
...
+            tar -cvf artifact.tar *
```
aminmarashi-binary: Why don't we compress the tar file? it needs `z` for `zip`.

aminmarashi-binary: We could also exclude `.git` directory here because the artifact won't use it anyway.

PR #1524, opened 2020-04-06
Refactor circleci config

afshinpaydar-binary:


Dockerfile
```diff
@@ -1,3 +1,3 @@
 FROM nginx:alpine
-COPY ./dist/compressed /usr/share/nginx/html/beta
```
4p00rv, line 2: this needs to dynamic based on type of release

afshinpaydar-binary: Dockerfile changed to dynamically load nginx location.

mahdipourziaei-binary:
The `npm_install` job runs `npm install` which reads from `package.json` so the cache key check is set up correctly, however, I think it would be better if we switch to use `npm ci` which reads from `package-lock.json`. `npm install` updates packages to the latest "wanted" version instead of the exact version that was tested, which might not be good practice for a staging/production release config. Otherwise, LGTM.

PR #1528, opened 2020-05-11
behnaz/oneHashtak-added

behnaz-fs:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "Webtrader",
-  "version": "2.2.8",
+  "version": "2.2.9",
```
mustofa-binary, line 3: same here 🙏 , we dont need to bump new version here

behnaz-fs: Sure, will change Webtrader version to 2.2.8

PR #1547, opened 2021-06-02
Mahdiyeh/ remove_cryptocurrency_market

mahdiyeh-fs:


matin-binary (approved):

matin-binary (changes requested):
The reason behind disabling cryptocurrencies is that **Crypto** only supports Multipliers in `contracts_for`, it is the same case for **Jump indices**. Since we don't have Multipliers in platforms other than Deriv-app, we are trying to hide based on the markets, but it is best to filter out based on shortcodes using regex to make sure it will handle other cases and avoid double work in future.
So I suggest using `\(?:JD[0-9])|(?:cry_[A-Z])\` to filter out shortcodes. `JD` for Jump indices and `cry` for Cryptos. 
Feel free to have a look into Zaki's changes in this PR https://redmine.deriv.cloud/issues/30826

matin-binary (changes requested):
Please help to update the PR

mahdiyeh-fs:
> regex

sure, I will work on that

src/navigation/menu.es6
```diff
@@ -50,8 +50,11 @@ export const extractChartableMarkets = (trading_times_data) => {
 };
 
 export const refreshMenu = (root, markets, callback) => {
+   const non_crypto = symbol => !/^(cry|JD)/i.test(symbol);
```
matin-binary, line 53: Please use a more descriptive name for this function!

mahdiyeh-fs: sure

matin-binary (approved):

PR #1553, opened 2021-06-27
Mahdiyeh/fix: missing_synthetic_indices_on_singapore_vpn

mahdiyeh-fs:


src/instruments/instruments.es6
```diff
@@ -8,57 +8,62 @@ import "common/util";
...
-                    submarket.instruments = _.map(symbols, function(sym) {
-                        active_symbols.push(sym.symbol);
-                        return {
-                            symbol: sym.symbol,
-                            display_name: sym.display_name,
-                        };
-                    });
-                    return submarket;
+      .cached
+      .authorize()
```
reza-fs, line 12: Shouldn't we check for logged in state before waiting on authorize? Since this requires a token to authorize a user and guest users don't have this. https://developers.binary.com/api/#authorize

mahdiyeh-fs: yeah, you're right, will be fixed

reza-fs (approved):

reza-fs (approved):

PR #1583, opened 2021-12-05
Mahdiyeh/Feature: Redirect non login uk/eu client

mahdiyeh-fs:


src/index.js
```diff
@@ -31,26 +31,100 @@ function loadAppId(callback) {
...
 }
 
+function getUrl() {
+    var server_url = localStorage.getItem('config.server_url') || 'frontend.binaryws.com';
+    return 'wss://' + server_url + '/websockets/v3';
+}
+
+function processRedirect(selected_language_name) {
+    loadAppId(function(err, app_id_json) {
+        var ip_country;
```
yashim-deriv, line 41:
```suggestion
        var client_country;
```
I think we can rename this as we are actually storing the client_country instead of the ip

src/index.js
```diff
@@ -31,26 +31,100 @@ function loadAppId(callback) {
...
+            var data = JSON.parse(response.data);
+
+            if(data.website_status){
+                ip_country = data.website_status.clients_country;
+                if (local_storage.get('oauth')) {
+                    var token = local_storage.get('oauth')[0].token;
+                    sendAuthorize(token);
+                } else {
+                    if(isEuCountrySelected(ip_country)) {
+                        window.location.href = getBinaryUrlWithoutLng('move-to-deriv');
```
yashim-deriv, line 84:
```suggestion
                        window.location.href = getBinaryUrlWithoutLng('move-to-deriv');
```
This can be extracted to its own function to make it clearer

yashim-deriv (commented):
Just a few reviews. The rest are good to go!

src/index.html
```diff
@@ -40,163 +40,177 @@
...
-                <b>Copy trading</b>
-                <p class="paragraph--br">
-                    Replicate the trades of more experienced traders in your own account.
-                </p>
-               <button onclick="openTradingPage()">Start trading</button>
+        <div id='close_banner_container' class='container invisible'>
+            <div class='close_banner_text'>
+                <img class='close_banner_img' src="v<version>/images/banner/icon_left.png"/>
+                <div class='close_banner_text_wrapper'>
+                    <h3>Binary.com is moving to Deriv on 30 November</h3>
```
matin-binary, line 98: 30 November is already passed, do we still need to include it?

matin-binary: also isn't it `30th` 😄

mahdiyeh-fs: actually there is no need for the banner since we redirect all of those clients to 'move-to-deriv', but as @prince said it should be just left like that

matin-binary (approved):

PR #1584, opened 2021-12-08
Mahdiyeh/Fix: duration_unit and contract_type for Oil/USD sybol

mahdiyeh-fs:


src/trade/tradeDialog.es6
```diff
@@ -543,8 +543,11 @@ function init_state(available,root, dialog, symbol, contracts_for_spot) {
       return { text, value: date };
     });
 
-    if (spot_starting_options) start_dates.unshift({ text: 'Now', value: 'now' });
-
+    if (spot_starting_options) {
+      start_dates.unshift({ text: 'Now', value: 'now' });
+    } else {
+      state.date_start.value = start_dates[0].value;
```
matin-binary, line 549: Don't you need to update `options` value, instead of manipulating state value?

mahdiyeh-fs: Since we use this value from state in other update methods, we need to update the state accordingly. 
For example in 'state.duration_unit.update' method we filter available based on 'start_type' which is depended to state.date_start.value and because the value is not updated before, it passes the empty array and cause the duration_unit bacome empty.

matin-binary (approved):

PR #1587, opened 2021-12-27
Mahdiyeh/Fix: undefined css in dialogExtend

mahdiyeh-fs:


src/cashier/withdraw.es6
```diff
@@ -39,7 +39,6 @@ class Withdraw {
             height: 400,
             'data-authorized': true,
             close: () => {
-                win.dialog('destroy');
```
matin-binary, line 42: does this leave us with a memory leak? or how do we make sure that the dialog is being destroyed?

mahdiyeh-fs: every time a dialog  get closed, it actually destroy it and when you click on according button, make a new dialog again.
in the withdraw dialog destroying the dialog occurs before restoring it, so cause an reference issue, so I remove the destroying to prevent this.

mahdiyeh-fs: I also add sanity check for it to make sure it does not lose it's functionality with this change

matin-binary: Thanks for the explanation, LGTM 👍🏼

matin-binary (approved):

src/cashier/withdraw.es6
mahdiyeh-fs: :bofa:

PR #1591, opened 2022-01-16
Mahdiyeh/Fix: currency on login

mahdiyeh-fs:


src/navigation/navigation.es6
```diff
@@ -159,7 +159,9 @@ const initLoginButton = (root) => {
 
       state.currency = data.authorize.currency;
       local_storage.set('currency', state.currency);
-      update_balance(data);
+      const loginId = local_storage.get('authorize').loginid;
+      if (data.authorize && data.authorize.loginid === loginId) update_balance(data);
```
salarhfz-fs, line 163: LGTM and I'm going to approve it but I've got 2 nitpicks.
- `loginId` or `login_id`? What's the project convention?
- Could we use `data.authorize?.loginid`?

salarhfz-fs (approved):

src/navigation/navigation.es6
mahdiyeh-fs: thanks salar for your fast and not furious approval :)
we do not have certain convention for variables in webtrader and since I used camelCase  before, I keep it like before.
also webtrader do not support optional chaining, so we could not use `data.authorize?.loginid?` .

salarhfz-fs: Thank you for the explanation :))

src/navigation/navigation.es6
```diff
@@ -159,7 +159,9 @@ const initLoginButton = (root) => {
 
       state.currency = data.authorize.currency;
       local_storage.set('currency', state.currency);
-      update_balance(data);
+      const loginId = local_storage.get('authorize').loginid;
```
mehrzad-fs, line 162:
```suggestion
      const loginId = local_storage.get('authorize')?.loginid;
```

src/navigation/navigation.es6
mehrzad-fs, line 163:
```suggestion
      if (data.authorize?.loginid === loginId) update_balance(data);
```

mahdiyeh-fs: @mehrzad-fs  jan as I explained above, the webtrader does not support the optional chaining, so we couldn't use what mentioned unfortunately :(

mehrzad-fs (approved):

matin-binary (approved):

PR #1604, opened 2022-05-16
Mahdiyeh/fix: fix destroyed dialog error on workplace

mahdiyeh-fs:
in the current version of webtrader we add destroyed dialog to the workplace table, so whenever we click on the table rows we get error which refers that this dialog is not initialized yet. In this PR the function `addDialog` is called on dialogOpen and `removeDialog` is called whenever the dialog get closed.

### DIFF
--- a/src/workspace/workspace.es6
+++ b/src/workspace/workspace.es6
@@ -267,9 +267,10 @@ export const init = (parent) => {
    parent.append(root);
    rv.bind(root[0], state);
 }
-export const addDialog = (name, clickCb, removeCb) => {
+export const addDialog = (name, id, clickCb, removeCb) => {
    const row = {
       name: name,
+      id: id,
       click: () => {
         manager_win && manager_win.dialog('close');
         clickCb();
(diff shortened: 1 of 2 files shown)


src/workspace/workspace.es6
```diff
@@ -267,9 +267,10 @@ export const init = (parent) => {
    parent.append(root);
    rv.bind(root[0], state);
 }
-export const addDialog = (name, clickCb, removeCb) => {
+export const addDialog = (name, id, clickCb, removeCb) => {
```
matin-binary, line 270: Why did you add `id` as the second argument to this function?
Are we passing the value once we invoke the function?

mahdiyeh-fs: we need this id for removeDialag function. yes we are.

matin-binary: Where do we invoke the function from?

mahdiyeh-fs: in windows.es6 file line 373

matin-binary (approved):

PR #1611, opened 2022-05-19
Include is not a function bug fixed

suisin-deriv:


matin-binary (approved):

src/common/util.js
```diff
@@ -429,7 +429,7 @@ function setupi18nTranslation(dict) {
       }
 
       function localize(node) {
-          if (node && node.className && node.className.includes('no-translation')) return;
+          if (node && node.className !== undefined && node.className.includes('no-translation')) return;
```
amina-deriv, line 432: can we just check `node?.className` here

suisin-deriv: I have tried node?.className but because of version issue so it couldn't be accepted when i try to build it.

amina-deriv (approved):

vijayasree-deriv (approved):

